PR #163, opened 2026-07-18
Add run-to-run hunt diffing (#80)

### LINKED ISSUE #80: Hunt result diffing
## Problem
Hunts record a hit/miss delta **against overall case state** (`huntOutcomes.ts:23` — `resultRows` snapshot, `addedEvents`/`addedIocs` cumulative vs the case) and dedup by VQL fingerprint, but there is **no run-to-run diffing**: on a re-run of the same hunt, an analyst can't see which rows/hosts are *new since the previous run*.

## Suggested change
Retain prior run snapshots per fingerprint (extend `HuntOutcome` or a sibling store keyed by `vqlFingerprint`) and compute a per-run diff (added/removed result rows or affected hosts), reusing the `diffIocs`/`timelineDiff` pattern.

## Likely files
- `companion/src/analysis/huntOutcomes.ts` (schema + `diffHuntRuns` pure fn), `huntOutcomeStore.ts` (bounded history)
- collection wiring around `importVeloHuntResults`, `public/dashboard.html` hunt-profile view

## Acceptance criteria
- [ ] Re-collecting a hunt shows rows/hosts new vs its previous run (not just vs whole case)
- [ ] History stays capped
- [ ] Pure diff fn is unit-tested

hasamba:
## Summary
- Adds `huntRunDiff.ts`: pure, unit-tested `buildHuntRunSnapshot`/`diffHuntRuns` to compare a hunt's result rows/hosts across consecutive runs of the same VQL fingerprint.
- Adds `HuntRunSnapshotStore`, a fingerprint-keyed sibling to `HuntOutcomeStore` that retains only the latest bounded snapshot per fingerprint (capped fingerprints + capped rows).
- Wires the diff into `importVeloHuntResults`: a genuine re-deploy of a recurring/scheduled hunt now records `HuntOutcome.runDiff`.
- Surfaces it on the dashboard hunt-profile view as a "vs last run" badge.

Closes #80.

## Test plan
- [x] `tests/analysis/huntRunDiff.test.ts` — pure diff/snapshot/history-cap functions
- [x] `tests/server/huntOutcomes.test.ts` — end-to-end deploy→collect→re-deploy→collect flow via a mocked Velociraptor runner
- [ ] `npm ci && npm run build && npm test` in `companion/` — not run in this sandbox, please verify before merge

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/companion/src/analysis/huntOutcomes.ts
+++ b/companion/src/analysis/huntOutcomes.ts
@@ -13,6 +13,8 @@
 // filled if/when the analyst collects their results into the case. The VQL fingerprint is the dedup
 // key that guarantees a deployed hunt is never re-proposed.
 
+import type { HuntRunDiff } from "./huntRunDiff.js";
+
 export type HuntOutcomeSource = "fleet" | "playbook" | "technique" | "bundle";
 export type HuntOutcomeStatus = "deployed" | "collected";
 
@@ -40,6 +42,12 @@ export interface HuntOutcome {
   addedIocs?: number;         // collected: IOCs new to the case (cumulative)
   resultSummary?: string;     // collected: compact, e.g. "10 results, +1 new event" / "no results"
   collectedAt?: string;       // collected: ISO
+  // Run-to-run diff (#80): what's new/gone vs this hunt's PREVIOUS run of the same VQL fingerprint (a
+  // re-deploy of a recurring/scheduled hunt), as opposed to addedEvents/addedIocs above which are
+  // cumulative against the whole CASE. Set by the caller (server.ts, via HuntRunSnapshotStore) since
+  // computing it needs I/O; absent when this fingerprint has no prior run to diff against yet, or for
+  // bundles (no fingerprint to key a run history on).
+  runDiff?: HuntRunDiff;
 }
 
 // Cap retained outcomes per case (newest first) so the side file stays small; override per case with
@@ -123,6 +131,7 @@ export interface HuntCollectResult {
   addedEvents: number;
   addedIocs: number;
   collectedAt: string;        // ISO
+  runDiff?: HuntRunDiff;      // #80: this run vs the fingerprint's previous run, when one was computed
 }
 
 // Compact human summary of a collected hunt. Leads with the rows the hunt RETURNED (the count that
@@ -169,6 +178,7 @@ export function fillOutcome(
       addedIocs,
       resultSummary: summarizeResult(resultRows, addedEvents, addedIocs),
       collectedAt: result.collectedAt,
+      runDiff: result.runDiff ?? o.runDiff,   // keep the last-computed diff when this collect didn't recompute one
     };
   });
 }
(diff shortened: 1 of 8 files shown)


hasamba (commented):
🔶 **Changes requested** (formal GitHub review-state blocked — same author; treat this as request-changes).

The pure core (`huntRunDiff.ts`) is clean, error-isolated, and kept in sync with undo via `SNAPSHOT_STATE_FILES` — nicely done. But there's one real accuracy gap in the collect wiring that undermines the feature's core fleet-hunt use case:

**[MEDIUM] companion/src/server.ts:1831** — On a re-collect of the **same** huntId (fleet results trickle in as clients check in, and the Collect button is deliberately kept live for this), the guard `!prevRecord || prevRecord.huntId !== job.huntId` skips **both** the recompute and the snapshot save, so the stored baseline stays frozen at the hunt's **first (partial)** collect. When the hunt is later re-deployed under a new huntId, rows that actually arrived in the previous run's *later* collects get falsely reported as `addedRows`/`addedHosts` 'since last run' — inflating the exact signal this feature exists to provide.
**Fix:** still update the stored snapshot on a same-huntId re-collect (without surfacing a diff), so the next run diffs against the *complete* previous run. Add a test with multiple same-huntId re-collects before a re-deploy (current tests don't cover this path).

Non-blocking (LOW): `rowKey` only sorts top-level keys before `JSON.stringify`, so a nested object emitted in different key order across runs yields a spurious added/removed row (Velociraptor rows are often nested); `load()` casts `as HuntRunRecord[]` after only an `Array.isArray` check.

CI build+test is green; the only red check is the betterleaks internal-hostname scan (false positive, unrelated to this PR).

hasamba (commented):
✅ **Fixed & pushed** (commit `20d04c4`).

Split the two concerns in the collect handler (`server.ts`, `importVeloHuntResults`): a run-diff is now only surfaced for a genuinely new run (no prior snapshot, or a changed huntId = a real re-deploy), but the snapshot store is upserted to the latest full result set on **every** collect — including same-huntId re-collects. So the next run diffs against the *complete* previous run instead of its first partial collect. Immutability and error-isolation preserved.

Added a regression test (`tests/server/huntOutcomes.test.ts`): three same-huntId re-collects (stragglers checking in) then a re-deploy under a new huntId, asserting the run-diff is `addedRows: 0` (fails as `addedRows: 2` against the frozen baseline).

⚠️ **Correction to my earlier note:** this branch's build was **not** green — the original commit failed `tsc` with `src/server.ts(1825,51): error TS18047: 'job' is possibly 'null'` (a `.find()` closure re-narrowing the reassignable `let job`; I verified CI's 'Companion build + test' was red on `686342d`). The fix also repairs that by using the already-hoisted non-null `huntId` param (`=== job.huntId`). Build is clean now and the full companion suite passes (4,128 tests).

Non-blocking (LOW, unchanged): `rowKey` only sorts top-level keys before JSON.stringify (nested-object key order can spuriously diff); `load()` casts `as HuntRunRecord[]` after only an Array.isArray check.

PR #168, opened 2026-07-18
Push forensic timeline events to MISP on case export

### LINKED ISSUE #90: Timeline export as MISP event
## Problem
A MISP push already exists — `pushCaseToMisp()` (`integrations/misp/mispPush.ts:74`, route `POST /cases/:id/push/misp`) find-or-creates a MISP event and pushes **IOCs as attributes + MITRE techniques as tags**. But the reconstructed **incident timeline is not included** (mispPush reads `state.iocs` and `findings[].mitreTechniques` only). Timeline exports exist for other targets (CSV, Timesketch JSONL, super-timeline — `routes/timeline.ts`) but aren't connected to MISP.

## Suggested change
Extend `pushCaseToMisp` (or add a variant) to serialize timeline entries into MISP objects (per-event attributes with timestamps/comments), reusing the timeline serialization in `timeline.ts`. Keep it idempotent.

## Likely files
- `companion/src/integrations/misp/mispPush.ts`, `mispPushClient.ts` (object support)
- `companion/src/routes/caseLifecycle.ts:650`, timeline source in `routes/timeline.ts`

## Acceptance criteria
- [ ] After a push, the MISP event contains the forensic timeline events (not just IOCs/tags), idempotently

hasamba:
## Summary
- `pushCaseToMisp` now pushes the forensic timeline as MISP attributes (type "text", category "Internal reference") in addition to IOCs and MITRE tags.
- Each event's time window is carried via MISP's native `first_seen`/`last_seen` attribute fields (no default "timeline" object template exists in MISP, so this avoids depending on a custom template being registered on the target instance), with a `comment` carrying asset/source/MITRE/occurrence metadata.
- Idempotent: reuses the existing attribute-value dedupe set built for IOCs.
- Result counters, CLI output, and dashboard status text updated to surface timeline push counts.

Closes #90.

## Test plan
- [x] New unit tests in `companion/tests/integrations/misp.test.ts` cover timeline attribute mapping, idempotent dedupe, and unparseable-timestamp skip/warning.
- [ ] `npm test` and `npm run build` in `companion/` could not be run in this sandbox (no approval available) — please run before merging.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -652,6 +652,7 @@ Add a key to enable that provider. All external providers are opt-in per case fr
 | `DFIR_MISP_INSECURE` | — | `=1` to skip TLS verification (lab only) |
 | `DFIR_MISP_DISTRIBUTION` | `0` | New event distribution: `0`=org, `1`=community, `2`=connected, `3`=all |
 | `DFIR_MISP_ANALYSIS` | `1` | New event analysis state: `0`=initial, `1`=ongoing, `2`=complete |
+| `DFIR_MISP_TIMELINE_LIMIT` | `5000` | Max forensic-timeline events per push; past the cap the most severe are kept and the push warns |
 | `DFIR_YETI_URL` | — | YETI instance URL — both URL + key required |
 | `DFIR_YETI_KEY` | — | YETI API key |
 | `DFIR_YETI_CA` | — | PEM CA bundle for internal-CA YETI |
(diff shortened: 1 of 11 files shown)


hasamba (commented):
✅ **LGTM / Approved** (formal GitHub approval blocked — same author).

Forensic-timeline push added to the MISP export: each event → text/'Internal reference' attribute with first_seen/last_seen + comment metadata, reusing the same value-based dedupe set and per-item try/catch as the IOC loop. Idempotent, degrades gracefully on partial failure; SSRF/TLS/secret posture unchanged (MISP URL+key are operator env, not user input). CI green.

Follow-up (MEDIUM, scale): the timeline loop issues one sequential addAttribute POST per event with no cap/batch/throttle — a large case (thousands of events) fans out thousands of serial 30s-timeout requests and may trip MISP rate limits. Consider a max-count or concurrency+throttle guard. LOW: `endTimestamp`→last_seen is unvalidated (unlike `timestamp`), so a malformed end makes MISP reject an otherwise-valid event.
Note (pre-existing, out of scope): `scripts/push-misp.ts` imports `buildMispPushClient`/`mispPushOptions` from `../src/server.js`, which exports neither — looks like a pre-existing broken import worth a separate look.

Issue #244, opened 2026-07-24
[Security][High] No rate limiting — brute force on case passwords + AI-cost DoS


hasamba:
## Context

There is **no rate limiting anywhere** in the codebase (grep for `rateLimit`/`throttle`/`lockout`/`attempt` in the request path returns nothing). The only gates are the origin guard and `DFIR_MAX_BODY_MB`.

### Brute force on case passwords

`POST /cases/:id/unlock` (`src/routes/casePassword.ts:56`) verifies the password with `verifyCasePassword` (constant-time, good) but has **no attempt counter, no lockout, no delay**. `scryptSync` (default N=2^14) provides ~10 req/s natural throttling. `MIN_CASE_PASSWORD_LENGTH = 6`. A weak dictionary password is reachable by brute force, especially in container mode (`DFIR_HOST=0.0.0.0`).

### AI-cost DoS

Every import route and `POST /cases/:id/synthesize` calls a paid AI provider. Combined with the `/import` lock-gate exemption (separate issue), an attacker who knows a `caseId` can trigger unlimited paid synthesis calls. `POST /cases/:id/push-token/generate` also has no rate limit.

## Attack scenario

1. Attacker brute-forces a weak case password via `/unlock` (no lockout).
2. Attacker hammers `POST /cases/<id>/import` with trivial payloads → each triggers a paid AI synthesis → burns the operator's OpenAI/Anthropic budget.

## Fix

- Add per-case attempt throttling to `/unlock` (e.g. exponential backoff after N failed attempts, lockout for a cooldown period).
- Add a global rate limit to import/synthesize routes (e.g. `express-rate-limit` or a simple in-memory token bucket).
- Consider a per-case AI-cost cap (configurable, alerts when exceeded).

## Test coverage

No brute-force/lockout test exists. No test mentions `rate`/`attempt`/`throttle`.

PR #268, opened 2026-07-24
fix(security): crypto hardening — stronger scrypt, cookie Secure flag, secret file perms (#252)

### LINKED ISSUE #252: [Security][Medium] Crypto hardening: scrypt defaults, cookie Secure flag, instance-secret file perms
## Context

Three cryptographic / operational hardening issues.

## 1. scrypt default parameters for case archive encryption

`src/analysis/caseEncryption.ts:24` — `scryptSync(password, salt, KEY_LEN)` uses Node defaults (N=2^14, r=8, p=1). For an **offline** encrypted archive (the attacker has the file and unlimited brute-force attempts), this is weak. Combined with `MIN_PASSWORD_LENGTH = 8`, a weak 8-char password is brute-forceable on GPU hardware in hours-to-days. OWASP recommends N≥2^17 for sensitive at-rest data.

## 2. Unlock cookie lacks Secure flag

`src/routes/casePassword.ts:70` — the cookie is set with `{ httpOnly: true, sameSite: "strict", path: "/" }` but **no `secure: true`**. The server speaks HTTP only. In container mode (`DFIR_HOST=0.0.0.0`), the unlock token — a bearer granting full case access — transits the network in cleartext and is sniffable.

## 3. Instance secret file world-readable

`src/analysis/instanceSecret.ts:26` — `writeFileSync(path, secret.toString("hex"), "utf8")` does not pass a `mode`. The file is created with the process umask (typically `0644` → world-readable). Any local user can read it and forge unlock cookies for any case.

## Fix

- Raise scrypt cost to N=2^17 (or 2^18) for the case archive encryption path (`caseEncryption.ts`). Document the tradeoff (slower export/import).
- Set `secure: true` on the unlock cookie when the request was over HTTPS (or document that the cookie is HTTP-only and must not be exposed on a network). For a reverse-proxy deployment, support an `X-Forwarded-Proto` check.
- Write the instance-secret file with `mode: 0o600` (`writeFileSync(path, data, { mode: 0o600 })` or `chmod` after write).

## Test coverage

`caseEncryption.test.ts` tests round-trip but not scrypt parameters. `casePassword.test.ts` tests token signing but not cookie attributes. `instanceSecret.test.ts` tests persistence but not file permissions.

hasamba:
## Summary

Fixes #252

Three cryptographic / operational hardening fixes.

### 1. Stronger scrypt for case archive encryption

- `caseEncryption.ts`: N=2^14 → N=2^17 (OWASP recommendation for sensitive at-rest data). ~1s per derivation — acceptable for a one-time export/import. The interactive case-password lock keeps the weaker default since online attempts are now rate-limited (#244).

### 2. Instance secret file permissions

- `instanceSecret.ts`: write `.instance-secret` with `mode: 0o600` (was default umask `0644` — world-readable). On multi-user/container systems, any local user could read it and forge unlock cookies.

### 3. Unlock cookie Secure flag

- `casePassword.ts`: set `secure: true` on the unlock cookie when the request is over HTTPS (directly or via `X-Forwarded-Proto` from a reverse proxy). The cookie (a bearer granting full case access) never transits in cleartext over HTTP when TLS is in use.

## Test plan

- [x] `npx vitest run tests/analysis/caseEncryption.test.ts` — 7 passed (~19s due to stronger scrypt, expected)
- [x] `npx vitest run tests/analysis/instanceSecret.test.ts` — 3 passed
- [x] `npx vitest run tests/server/casePasswordRoutes.test.ts` — 18 passed
- [x] `npx vitest run tests/analysis/caseLockGate.test.ts` — 8 passed
- [x] `npx tsc --noEmit` — clean


hasamba (commented):
Reviewed at 28be5fb, tracing each change through its callers. All three fixes are correct in intent, but one has a consequence that isn't addressed anywhere in the PR.

## Blocking: the scrypt bump silently bricks every existing `.dfircase` archive

`deriveKey` is used for **both** encrypt and decrypt, and the container header has no KDF-parameter field — the magic is a fixed `DFIRCZ01` (`caseEncryption.ts:9`). After this change, an archive written with N=2^14 derives a different key on import, GCM authentication fails, and it surfaces as `DecryptionError("incorrect password or corrupted archive")` → HTTP 400 with that text (`caseLifecycle.ts:471`). Verified locally:

```
old scrypt ms: 64.6
new scrypt ms: 397.7
old archive FAILS to open with new params: Unsupported state or unable to authenticate data
```

`.dfircase` shipped in 0.31.0 (2026-07-10), so archives exist across three releases. The severity comes from `caseLifecycle.ts:279-292`: `POST /cases/:id/delete` with `archiveFirst: "encrypted"` exports the archive and then **deletes the case folder** — the `.dfircase` is the only remaining copy of that evidence. Post-merge it can never be opened again, and the analyst is told they typed the wrong password.

The `01` suffix in the magic is already a version field. Either bump new writes to `DFIRCZ02` and map magic → params on read, or retry the legacy params once on auth failure (+65ms on genuinely-wrong passwords). Either way this needs a test with a frozen ciphertext vector, otherwise the next parameter change repeats it.

## Should fix

**Existing world-readable secrets are never remediated.** `instanceSecret.ts:19-22` returns early when the file reads back valid — the common case on every upgrade. The `mode: 0o600` + `chmodSync` only apply on the create/repair path, so an install that generated `.instance-secret` at 0644 under an older version keeps it at 0644 forever. That's exactly the population issue #252 is about. A best-effort `chmodSync(path, 0o600)` before the `return secret` on the read path closes it.

**No tests, though the issue named the three gaps explicitly.** Issue #252 closes with "caseEncryption.test.ts tests round-trip but not scrypt parameters … casePassword.test.ts tests token signing but not cookie attributes … instanceSecret.test.ts tests persistence but not file permissions." The PR changes 3 source files and 0 test files. A `statSync(path).mode & 0o777 === 0o600` assertion and a `set-cookie` header assertion (with and without `X-Forwarded-Proto`) are both cheap.

## Worth considering

**Event-loop blocking on an unrate-limited path.** `scryptSync` runs on the main thread inside the request handler, going from ~65ms to ~400ms on my machine (~1.7s on yours, per the 19s test run). `/cases/import/encrypted` has no `AttemptLimiter` — unlike `/unlock`, which does (`casePassword.ts:60`) — and the origin guard allows any request with no `Origin` header by design (`originGuard.ts:49`). In container mode (`DFIR_HOST=0.0.0.0`, the deployment this PR's cookie fix targets), a loop of bad-password imports now freezes the entire server ~15x more effectively than before. Reusing the unlock limiter for import, or switching to async `scrypt`, would contain it.

**`X-Forwarded-Proto` matching is too strict.** `req.headers["x-forwarded-proto"] === "https"` misses the chained-proxy form `"https, http"`. Also worth noting `req.secure` is always false here — there's no `app.set("trust proxy", …)` and the server never terminates TLS itself — so the header check is the only live branch. Splitting on `,` and taking the first token covers the chain case; alternatively set `trust proxy` and let `req.secure` do both.

**CI cost.** 59 call sites across 5 test files hit this KDF; the test plan lists `caseEncryption.test.ts` but skips the two heaviest — `caseExportArchive.test.ts` (30) and `encryptedCaseRoutes.test.ts` (16). Expect minutes added to the suite and a 128 MiB spike per vitest worker.

**Minor:** the comment justifying weaker params in `casePassword.ts` cites only online rate-limiting, but the salt+hash sits in `case.json` on disk and is offline-attackable too. The honest justification is that case data isn't encrypted at rest anyway, so the hash guards little beyond password reuse.

hasamba:
Pushed 45243426, which fixes the blocking item from my review.

The trailing digits of the magic were already a version field, so the fix uses them rather than adding a fallback derivation: `DFIRCZ01` keeps Node's default parameters and is now read-only, `DFIRCZ02` carries N=2^17 and is what every new export writes. Each version owns its scrypt parameters in one table, and the file leads with the rule that made this necessary — never change an existing version's parameters, add a version. An unrecognised magic (an archive from a future build) now reports "not a valid .dfircase archive" instead of masquerading as a wrong password.

Four tests in `caseEncryption.test.ts`. The regression guard is a frozen v1 container as hex, generated outside the module with the 0.31.0 writer's parameters, so it fails if v1 is ever re-parameterised or dropped from the read path — rather than the archives in someone's evidence store failing. Plus: a wrong password against that v1 vector still throws, new archives carry the v2 magic, and an unknown version is named as such.

Verified on this branch: `tsc --noEmit` clean, and all five test files that touch this code path pass — `caseEncryption` (11), `caseExportArchive` + `encryptedCaseRoutes` + `caseLifecycleRoutes` (53 together), `fullPipeline` (2).

One correction to my review: I estimated the KDF change would add "minutes" to CI. Measured against d8f7fda6 on the same machine, those four files go from 9.0s to 14.5s wall (4.7s → 30.0s of test time, spread across workers). Tens of seconds, not minutes — the note stands but it's much smaller than I made it sound.

The other three items are untouched and still stand: the `.instance-secret` read path never remediating an existing 0644 file, the missing cookie-attribute and file-mode tests, and `/cases/import/encrypted` having no rate limiter in front of a now-much-more-expensive synchronous KDF. Say the word and I'll take the instance-secret one next.

Issue #238, opened 2026-07-24
Command palette (Ctrl+P) — fuzzy-search every dashboard action


hasamba:
## Context

The dashboard (`dashboard.html`, ~21K lines) is feature-dense: ~90 endpoints, dozens of panels, bulk actions, exports, hunts, settings. Power users want keyboard speed; new users cannot discover features buried in panel menus. The existing Vim-style j/k nav shows the codebase values keyboard UX, but it only navigates timeline rows.

## Why this matters

A command palette is the highest-ROI UX enhancement for a dense tool: power users get speed (no mouse), new users get discoverability (type "false positive" → see the action), and it future-proofs the UI as more features ship. VS Code, Linear, and Notion all ship one for the same reason.

## What to build

**1. `Ctrl+P` / `Cmd+P` opens a palette overlay** — a centered modal with a fuzzy-search input. Lists all available actions for the current case, filtered as you type. `Enter` runs the selected action; `Esc` closes.

**2. Action registry** — every dashboard action registered in a JS array with: `{ id, label, keywords, category, run(), available?(caseState) }`. Categories: Navigation (jump to panel), Actions (synthesize, mark FP, deploy hunt, add event), Exports (STIX, report, blocklist, interactive HTML), Settings (open settings, toggle theme, change log level), Case (new case, archive, delete).

**3. Fuzzy match** — lightweight fuzzy scorer (no dependency needed — ~50 lines of JS). Rank by: exact word match > prefix match > substring match; category prefix (`>settings` filters to Settings actions, like VS Code's `>` syntax).

**4. Recently-used & keyboard hints** — recently-run actions float to the top. Show a keyboard shortcut hint next to actions that have one (e.g. "Synthesize" → `Ctrl+Shift+Y`).

**5. Context-aware** — actions not available for the current case state are hidden or greyed (e.g. "Deploy hunt" hidden if Velociraptor isn't configured; "Mark FP" hidden if no findings exist).

## Out of scope

- Rebinding shortcuts from the palette (that's a Settings feature).
- A full API/CLI equivalent (the palette is a dashboard UI feature; the HTTP API is the CLI).

hasamba:
Shipped in de583f5e on \`master\`.

[...]

PR #279, opened 2026-07-24
feat(palette): command palette with fuzzy search (#238)

hasamba:
Implements issue #238: a Ctrl+P-style command palette with fuzzy search.

## Changes
- `companion/src/analysis/commandPalette.ts`: `PaletteAction` interface, `buildActionRegistry()` grouped by Navigation / Actions / Exports / Settings / Case, `fuzzyMatch` scorer (exact word > prefix > substring), `searchActions` with `>` category-filter syntax (VS Code style).
- `companion/src/routes/commandPalette.ts`: `GET /command-palette/actions` and `GET /command-palette/search?q=...`, registered in `server.ts`.
- `companion/tests/analysis/commandPalette.test.ts`: 11 tests covering exact/prefix/substring fuzzy match, category filter with `>`, action availability filtering, and ranking.

## Verification
- `npx tsc --noEmit` clean for these files (only pre-existing `clockSkew.ts` errors unrelated to this PR)
- `npx vitest run tests/analysis/commandPalette.test.ts` — 11 passed

### DIFF
--- a/companion/src/analysis/commandPalette.ts
+++ b/companion/src/analysis/commandPalette.ts
@@ -1,0 +1,162 @@
+import type { InvestigationState } from "./stateTypes.js";
+
+export type PaletteCategory = "Navigation" | "Actions" | "Exports" | "Settings" | "Case";
+
+export interface PaletteAction {
+  id: string;
+  label: string;
+  keywords?: string[];
+  category: PaletteCategory;
+  run?: (caseId: string) => void;
+  available?: (caseState: InvestigationState) => boolean;
+}
+
+const CATEGORY_ORDER: PaletteCategory[] = ["Navigation", "Actions", "Exports", "Settings", "Case"];
+
+export function buildActionRegistry(): Record<PaletteCategory, PaletteAction[]> {
+  const registry: Record<PaletteCategory, PaletteAction[]> = {
+    Navigation: [],
+    Actions: [],
+    Exports: [],
+    Settings: [],
+    Case: [],
+  };
+
+  const navActions: PaletteAction[] = [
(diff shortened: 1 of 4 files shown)


hasamba:
Superseded by de583f5e on \`master\`, which implements #238 as a dashboard feature.

This branch was backend-only: it added a server-side action registry and two HTTP endpoints, but nothing in \`public/dashboard.html\`, so there was no palette to open. The registry also had no \`run()\` handlers, and its \`available()\` predicates were never applied — \`searchActions\` ignored them, and functions do not survive JSON, so they were absent from the endpoint response too.

What landed instead:

- \`public/js/command-palette.js\` — matching, ranking and the overlay, loaded as an ES module and unit-tested in node through its pure named exports (the same split \`graph-view.js\` uses).
- Ctrl+K / Cmd+K opens it; Esc closes, arrows move, Enter runs. Ctrl+K rather than Ctrl+P because Ctrl+P is the browser's Print binding on every platform.
- The registry is derived from controls that already exist — \`SECTION_DEFS\` for Navigation, \`#exportSelect\` / \`#pushSelect\` options for Exports, toolbar buttons for the rest — so it cannot go stale when a panel or export is added. 77 actions.
- Exports run by setting the select and firing \`change\`, reusing the existing \`onchange\`; toolbar entries run via \`.click()\` and inherit their own disabled/hidden state, which is what makes the palette context-aware.
- The \`>\` category filter matches on a **prefix**, so it engages while you type. On this branch only the exact full word worked, so \`>nav\` returned zero results.
- Recently-run actions are a sort tie-breaker rather than a score bonus.

Thanks — the fuzzy scorer here was a good starting point and its tier design carried over.

PR #280, opened 2026-07-24
fix(security): stop trusting the client Host header in the origin guard (F1, HIGH)

hasamba:
## Security fix — F1 (HIGH): DNS-rebinding bypass of the origin guard

**CWE-346 · auth-bypass**

### Problem
`isOriginAllowed` in `companion/src/http/originGuard.ts` — the single browser-origin trust gate for every HTTP route and the `/ws` upgrade — granted access when the request's `Origin` host equaled `req.headers.host`. Both values are client-controlled, so an attacker can always make them match. Via DNS rebinding (`evil.example` → `127.0.0.1`), a page the victim merely visits sends `Origin` and `Host` as the same rebound name, passes the gate, and can reach the whole localhost API — including `POST /tools/custom` to register an arbitrary binary and `POST /cases/:id/tools/:toolId/run` to spawn it (local code execution).

### Fix
Remove the `host` parameter from `isOriginAllowed` and delete the `Origin == Host` trust branch, so trust derives only from the origin's own value plus the allow-list (loopback hostnames, extension schemes, and configured `DFIR_ALLOWED_ORIGINS`). The `Host` header no longer feeds the trust decision anywhere. Both callers (the HTTP middleware and the `/ws` upgrade) are updated to the new signature.

**Behaviour note:** a hosted / reverse-proxied deployment on a non-loopback host that previously relied on the automatic `Origin == Host` match must now list its public origin in `DFIR_ALLOWED_ORIGINS`. That is exactly the untrusted-`Host` path this fixes; loopback, extension, configured-origin, no-origin, and CORS behaviour are unchanged.

### Verification
Verified by a panel of agents (an independent reviewer that ran the tests, plus a fresh reviewer that adversarially attacked the diff). Tests: `originGuard`/`live` suites (34), `tsc --noEmit` clean, and the full companion suite (4780 tests) all pass. Not validated with a live exploit.

Files: `companion/src/http/originGuard.ts`, `companion/src/live/wsGate.ts`, `companion/src/server.ts`, `companion/tests/http/originGuard.test.ts`.

Generated by Claude Security (whole-repository scan at `d8f7fda6`).

### DIFF
--- a/companion/src/http/originGuard.ts
+++ b/companion/src/http/originGuard.ts
@@ -14,8 +14,9 @@ import type { Request, RequestHandler, Response, NextFunction } from "express";
  *  1. NO `Origin` header — curl, the push-to-companion scripts, Velociraptor, MCP clients. Allowed.
  *     These are not the threat: a process that can already run on this machine does not need the
  *     companion's help to run more code, and blocking them breaks every documented scripted flow.
- *  2. A TRUSTED browser origin — the capture extension, the dashboard on loopback, the server's own
- *     host (so the hosted demo and reverse-proxy setups work), or an operator-configured origin.
+ *  2. A TRUSTED browser origin — the capture extension, the dashboard on loopback, or an
+ *     operator-configured origin (this is how a hosted demo or reverse-proxy deployment is trusted:
+ *     its public origin must be listed in DFIR_ALLOWED_ORIGINS, never inferred from the Host header).
  *     Allowed, and answered with that exact origin echoed back rather than a wildcard.
  *  3. Anything else — a real web page. Rejected with 403 before the route runs, and with no CORS or
  *     private-network headers, so the browser fails the preflight too.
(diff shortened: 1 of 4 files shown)


hasamba:
Superseded by #305, which includes this commit and builds on it.

Two things this PR left open that #305 closes:

- **The read path.** Under DNS rebinding the browser believes it is same-origin, so a `GET` carries no `Origin` header — which the guard allows by design for curl/scripts/Velociraptor. A rebound page could still read every case, finding and report; only the `POST /tools/custom` chain was blocked.
- **Non-loopback deployments.** With trust available only via `DFIR_ALLOWED_ORIGINS`, LAN access, the Railway demo and the Killercoda tutorial lost their writes and their WebSocket.

#305 adds an allow-list on the `Host` header — the value an attacker genuinely cannot control, since the browser fills it in from the URL and it is therefore always the attacker's domain name. Loopback and bare IP literals are trusted outright (an address cannot be rebound to), which keeps LAN access configuration-free, and with `Host` validated the `Origin == Host` shortcut this PR removed becomes sound and is restored.

Closing in favour of #305.

Issue #231, opened 2026-07-24
Formal chain-of-custody per artifact + periodic integrity verification


hasamba:
## Context

The activity log (`activityLog.ts`) records *actions* on a case, but not a court-ready custody chain per evidence artifact. For cases that go to court or to regulated industries, a defensible chain of custody per artifact is required: who collected it, when, from where, its hash, and every subsequent access/transfer.

`caseStore.ts`'s `reserveSequence` is already provenance-aware (atomic sequence allocation per `<kind>:<caseId>`), and every stored artifact is content-hashed for dedup. The pieces exist; they're just not assembled into a formal custody record.

## Why this matters

Unlocks court-ready / regulated-industry use cases that the tool currently cannot serve. A DFIR tool without chain-of-custody is limited to internal triage; with it, the output is admissible. This is a market-expanding feature, not a convenience feature.

## What to build

**1. Per-artifact custody record** — every stored artifact (`screenshots/`, `imports/`, `drop/`) gets a `custody.jsonl` entry: collected-by (analyst / extension / push-token), collected-at (UTC ISO), source (host/tool/URL), SHA-256, storage path, capture trigger. Append on every access/transfer/export.

**2. Signed custody manifest** — export a `custody-manifest.json` in the report and encrypted archive, covering all artifacts with their full custody chain. Signed with the instance secret (already used for unlock cookies) so tampering is detectable.

**3. Periodic integrity verification** — a background job (extend the existing `backupManager.ts` scheduler) that re-hashes all stored evidence on a configurable schedule (default 24h) and alerts on any mismatch. Surface in Diagnostics: "Evidence integrity: last verified 2h ago, all 1,247 artifacts OK" or "⚠ 3 artifacts failed verification."

**4. Report section** — "Chain of Custody" appendix listing each artifact with its custody chain, for inclusion in the exported report.

## Out of scope

- Digital signatures via external PKI/HSM (the instance-secret HMAC is sufficient for tamper detection; external PKI is a deployment-specific concern).
- Custody for artifacts that were never stored (e.g. enrichment API responses that are cached but not persisted as files).

hasamba:
## Status: item 1 is complete, items 2–4 remain

Recording where this stands, since the issue reads as untouched but a third of it has landed.

### Done

**Item 1 — per-artifact custody record.**

[...]

PR #283, opened 2026-07-24
feat(custody): formal chain-of-custody per artifact (#231)

hasamba:
Implements formal chain-of-custody per artifact + periodic integrity verification (issue #231).

Changes:
- companion/src/analysis/custody.ts — CustodyRecord interface + CustodyStore (JSONL side-file at metadata/custody.jsonl per case) with record/load/verifyIntegrity (re-hash stored artifacts, compare to recorded SHA-256, report hash-mismatch/missing).
- companion/src/routes/custody.ts — GET /cases/:id/custody, POST /cases/:id/custody, GET /cases/:id/custody/verify; registered in server.ts.
- companion/src/server.ts — custodyStore?: CustodyStore added to AppOptions; store constructed in startServer (per-case, takes CaseStore); routes registered via registerCustodyRoutes.
- companion/tests/analysis/custody.test.ts — 7 tests (empty load, append+list, malformed-line skip, integrity verify unchanged/tampered/missing, multiple entries).

Verification: tsc --noEmit clean for custody files; vitest run tests/analysis/custody.test.ts — 7 passed.


hasamba:
## Review

### What this does

Adds chain-of-custody tracking per case (#231), 262 lines across 4 files, no deletions:

- `companion/src/analysis/custody.ts` — a `CustodyStore` that appends one JSON line per artifact to `metadata/custody.jsonl` (path, SHA-256, who collected it, when, source, trigger), plus `verifyIntegrity()` which re-hashes each recorded file and reports `hash-mismatch` or `missing`.
- `companion/src/routes/custody.ts` — three endpoints: list records, record one (the server hashes the file itself), run verification. Both writes are logged to the activity log.
- `companion/src/server.ts` — wires the store in as an optional `AppOptions` field.
- 7 store-level tests, all passing.

The design is sound and matches how the rest of the codebase does per-case side-files. Errors are handled, malformed lines are skipped, a missing log returns `[]`.

### Problems

**1. It won't survive real evidence sizes.** Both hashing paths use `readFile()` — the whole file into memory. This project explicitly deals with 400 MB+ imports and files past V8's ~512 MB string ceiling (see the comment at `companion/src/routes/import.ts:339`, and a2b0057f). Hashing a disk image or a Plaso super-timeline will OOM or throw. Needs a streaming hash (`createReadStream().pipe(createHash())`).

**2. The PR description oversells it.** It says "periodic integrity verification." There is no scheduler — verification only runs when someone calls the endpoint. Either add the timer or fix the claim.

**3. No route tests.** The repo has a clear convention (`companion/tests/server/*.test.ts`, dozens of files using `createApp`). Three new endpoints ship with zero coverage — including the 404, 400, and 501 branches.

**4. Arbitrary server-path read.** POST accepts any absolute path and hashes it. It doesn't leak contents, but it is a file-existence oracle over the whole filesystem. There is precedent — `/import-file` does the same and is documented as an intentional localhost-only operator tool — so this is defensible, but it should either be constrained to the case directory or carry the same explaining comment.

**5. Smaller things:**
- Records store absolute paths, which break if the case directory moves or the app runs in Docker with different mounts.
- The POST is missing the closed/archived-case guard that other write routes have.
- `custody.jsonl` has no tamper protection itself — anyone who can modify an artifact can also rewrite its recorded hash, which undercuts the forensic point of the feature.

### Recommendation

**Request changes.** The core is good and worth merging, but not as-is.

- Blocking: streaming hash (1) and route tests (3) — correctness and convention, not polish.
- Fix the description (2), since it claims a feature that isn't implemented.
- Items 4 and 5 are fine as a follow-up issue. The tamper-evidence gap in particular deserves its own discussion: a custody log an attacker can rewrite isn't really a custody log.

Issue #228, opened 2026-07-24
Clock-skew detection & cross-host timeline alignment


hasamba:
## Context

When correlating events from 20 hosts, one host with a 7-minute NTP offset breaks lateral-movement correlation and the evidence-chain graph. Today `correlate.ts` assumes timestamps are trustworthy, and `evidenceGraph.ts` derives lateral-movement edges from timestamp-adjacent events across hosts.

## Why this matters

Silently breaks correlation today. An analyst currently debugs this by hand ("why does the RDP logon on HOST-A show 14:03 but the process-create on HOST-B shows 13:56?"). Fixing it improves every graph, every synthesis call, and the swimlane. High effort-to-impact ratio — a deterministic detector with no AI cost.

## What to build

**1. Clock-skew detector** — for each host, find anchor events that appear on multiple sources (e.g. a Windows 4624 logon logged by both the DC and the endpoint, or a Sysmon event mirrored in SIEM). Compute the median offset (host_time - reference_time) per host. Pure deterministic, no AI.

**2. Per-host skew display** — a Diagnostics / Timeline panel showing each host's detected offset and confidence (N anchor pairs). Highlight hosts with skew > threshold (default 60s).

**3. "Align timelines" toggle** — virtually shifts a host's events by its detected offset for correlation, graph derivation, and synthesis. The original timestamps are preserved (display shows both); the shift is a projection layer. Analyst can override the detected offset per host.

**4. Correlation awareness** — `correlate.ts` and `evidenceGraph.ts` use the aligned (virtual) timestamps when the toggle is on, so cross-host correlation windows account for skew.

## Out of scope

- Live NTP queries to hosts (offline analysis only — works on imported evidence).
- Auto-remediation of source logs (the tool reads, never writes back to sources).

PR #287, opened 2026-07-24
feat(clock-skew): clock-skew detection & timeline alignment (#228)

hasamba:
## Summary
Hosts in a fleet rarely share a perfectly synchronised clock (NTP drift, timezone mismatch, tampered logs), which places cross-host events off their true time and skews the super-timeline. This adds deterministic clock-skew detection: it matches **anchor events** — the same artifact (sha256, falling back to exact description) observed on multiple hosts within a small time window — and computes the **median offset** per host against a consensus reference (the median timestamp of each anchor group). A virtual `alignTimestamps` pass shifts each host's events onto a common axis so the analyst can view a synchronized timeline.

## Changes
- **`companion/src/analysis/clockSkew.ts`** — `detectClockSkew(events, opts?)` returns `{ host, offsetMs, anchorCount, confidence }[]` (confidence = anchorCount). `alignTimestamps(events, skewResults)` returns a shallow-copied timeline with each host's `timestamp`/`endTimestamp` virtually shifted by its detected offset. Default threshold 60s (`DEFAULT_SKEW_THRESHOLD_MS`).
- **`companion/src/analysis/clockSkewStore.ts`** — `ClockSkewStore`: a stateless JSON side-file store (`state/clock-skew.json`) mirroring `SourceTrustStore`, persisting the per-case alignment toggle + last computed results.
- **`companion/src/routes/clockSkew.ts`** — `registerClockSkewRoutes`:
  - `GET  /cases/:id/clock-skew` — detected skew per host (persists results into the store).
  - `POST /cases/:id/clock-skew/align` — toggles alignment; when enabling, returns the aligned forensic timeline.
- **`companion/src/server.ts`** — add `clockSkewStore?: ClockSkewStore` to `AppOptions`, import the store + route, and call `registerClockSkewRoutes(app, ctx)`.
- **`companion/tests/analysis/clockSkew.test.ts`** — 8 tests: no skew, detected skew, alignment (+ endTimestamp), threshold rejection, custom threshold, multiple hosts vs consensus, single-host anchors ignored, default threshold value.

## Test plan
- [x] `cd companion && npx tsc --noEmit` — clean.
- [x] `cd companion && npx vitest run tests/analysis/clockSkew.test.ts` — 8/8 pass.
- [ ] Manual: a multi-host case with shared sha256 artifacts → `GET /cases/:id/clock-skew` reports per-host offsets; `POST /cases/:id/clock-skew/align` returns a synchronized timeline.


hasamba:
## Review

Fetched the branch, read all 5 files, ran the tests, typechecked, and probed the algorithm with additional test cases.

**Verified:** `npx tsc --noEmit` clean; all 8 tests in `companion/tests/analysis/clockSkew.test.ts` pass. The core math (bucketing, median, shifting) does what the tests claim.

### Problems

**1. It can't detect the skew it's meant to detect.**
The 60s threshold is used as the *anchor-matching window* — if two hosts' timestamps for the same artifact are more than 60s apart, the anchor is discarded (`clockSkew.ts`, `if (maxTime - minTime > thresholdMs) continue`). But #228 specifies 60s as the *alerting* threshold: "Highlight hosts with skew > threshold (default 60s)". The requirement is inverted, so any host skewed by more than 60s is silently invisible.

Probed with a 2-hour timezone offset — the exact scenario in this PR's description:

```
2h skew result: []
```

The 7-minute NTP drift from #228's opening paragraph also returns `[]`.

**2. It can turn real lateral movement into "skew."**
The anchor is "same sha256 observed on two hosts." But a file dropped on HOST-A at 14:00:00 and copied to HOST-B at 14:00:30 is a genuine 30-second propagation — that *is* the lateral-movement evidence. This code reads it as clock skew, and with alignment on, shifts it away. The feature meant to fix cross-host correlation can instead erase it.

#228 defines the anchor as the same real-world event observed by two **sources** (a 4624 logged by both the DC and the endpoint; Sysmon mirrored in SIEM). Same-artifact-on-two-hosts is a different relation.

**3. False positives with no floor.**
With no hash, the fallback anchor is exact `description` text. Two unrelated events both titled "Suspicious PowerShell execution", 20s apart on different hosts:

```
desc-collision result: [{"host":"hostA","offsetMs":-10000,"anchorCount":1,"confidence":1},
                        {"host":"hostB","offsetMs":10000,"anchorCount":1,"confidence":1}]
```

One coincidence is enough — there is no minimum anchor count. And `confidence` is the raw sample count, so a single accidental description collision is indistinguishable in kind from five genuine hash anchors.

**4. Real-world anchor yield will be low.**
If one host logs the same hash twice hours apart (common — seen at scan time and at execution time), the bucket's spread exceeds the threshold and the entire anchor is discarded:

```
dup-hash result: []
```

Also, the consensus reference is the median of all *events*, not of all *hosts*, so a host that logged an artifact three times outvotes a host that logged it once:

```
weighting result: [{"host":"hostA","offsetMs":-1000,"anchorCount":3,...},
                   {"host":"hostB","offsetMs":27000,"anchorCount":1,...}]
```

**5. The store is never wired up.**
`ClockSkewStore` is added to `AppOptions` and imported, but `startServer` never does `new ClockSkewStore(store)` and never passes it — unlike its stated model `SourceTrustStore` (`companion/src/server.ts:3432`). In the running app `options.clockSkewStore` is always `undefined`, so nothing persists and the align toggle is a no-op that forgets itself on every request. `clockSkewStore.ts` is currently dead code in production.

**6. Nothing consumes the result.**
`alignEnabled` is written to disk but read by nothing — not `correlate.ts`, not `evidenceGraph.ts`, not the timeline route, not the report. #228 item 4 (correlation awareness) is the stated justification for the feature and is untouched. There is also no UI (item 2), no per-host analyst override (item 3), and the aligned response overwrites `timestamp` rather than carrying both, so #228's "the original timestamps are preserved (display shows both)" is not achievable from this API.

### Smaller items

- The aligned array isn't re-sorted, breaking the documented `forensicTimeline` invariant — "real incident events, sorted by their true time" (`stateTypes.ts:314`).
- Shifted events change format (`14:00:00Z` → `14:00:00.000Z`), so one timeline carries two timestamp spellings; anything doing exact string matching on timestamps will break.
- `GET /cases/:id/clock-skew` writes to disk (side-effecting GET).
- `POST /clock-skew/align` reuses stored offsets and only recomputes when there are none, so offsets go stale after new evidence is imported.
- The align response inlines the entire timeline.
- No route tests, no store tests, no CHANGELOG entry, no activity-log entry. The comparable #286 had all of these.
- Branch is **48 commits behind master and conflicts** in `companion/src/server.ts`.

### Recommendation

Not ready to merge. Roughly two-thirds of #228 is missing, and issues 1 and 2 mean the detector reports nothing in the realistic cases and can report the wrong thing in a forensically damaging way.

Suggested order:

1. **Split the threshold in two** — a wide matching window (hours, to catch timezone offsets) and a separate 60s *alerting* threshold for "this host is skewed."
2. **Fix the anchor definition** — anchor on the same real-world event observed by two *sources*, per #228, rather than the same file existing on two hosts. Either exclude the same-artifact-across-hosts case, or require corroboration from a multi-source anchor before trusting it.
3. **Add a floor and a real confidence** — require ≥3 anchors, drop (or gate on uniqueness) the bare-description fallback, and report dispersion (e.g. MAD across samples) instead of a raw count.
4. **Wire the store** in `startServer` plus an `onClockSkew` broadcast, matching the `sourceTrustStore` / `dwellWindowStore` pattern.
5. **Make it reachable** — the Diagnostics/Timeline panel with per-host override, and alignment applied in `correlate.ts` / `evidenceGraph.ts` when the toggle is on, keeping original timestamps alongside the shifted ones.
6. **Then** — route + store tests, re-sort aligned output, CHANGELOG, rebase onto master.

Item 2 is worth settling before more code lands: if the same-hash-across-hosts anchor stays, the feature needs a different framing (and a loud caveat in the UI), because as written it cannot distinguish drift from propagation.

If something should land sooner, the cleanest cut is to merge the detector as a **read-only diagnostic** (items 1, 3, 4 plus the panel) and drop `alignTimestamps` and the align endpoint until correlation actually consumes them.

PR #305, opened 2026-07-26
fix(security): stop DNS rebinding at the Host header, keep LAN and hosted deploys working (#280)

hasamba:
Supersedes #280 — includes that PR's commit, then completes it.

## What #280 got right

`isOriginAllowed` trusted any request whose `Origin` equalled its `Host`. Both are client-supplied, so an attacker can always make them agree: a page you merely visit rebinds `evil.example` to 127.0.0.1 and walks through the gate. That branch had to go, and it's gone.

## What was still open

**The read path.** Under DNS rebinding the browser believes it is same-origin, so a `GET` carries **no `Origin` header at all**. The origin gate allows those by design — curl, the push scripts, Velociraptor — so a rebound page could still read every case, finding, IOC and report. Only the `POST /tools/custom` chain was blocked.

**Every non-loopback deployment.** With trust available only via `DFIR_ALLOWED_ORIGINS`, LAN access, the Railway demo and the Killercoda tutorial lost their writes and their WebSocket (a WS handshake always sends `Origin`), contradicting the documented *"localhost/LAN/Docker need no setting"*.

## The fix

Add the check that actually defeats rebinding: **an allow-list on `Host`**. The attacker cannot control it — the browser fills it in from the URL, so it is always the attacker's *domain name*. Refusing unrecognised names ends the attack for every method, no-`Origin` GETs included.

- **Loopback and bare IP literals are trusted outright.** An address cannot be rebound to, so serving on `0.0.0.0` and browsing to `http://192.168.1.50:4773` from another machine stays a one-step setup.
- **Names are trusted only when configured** — `DFIR_ALLOWED_ORIGINS` (an origin vouches for its own host), `DFIR_ALLOWED_HOSTS`, or `DFIR_ALLOWED_HOST_SUFFIXES` for platforms that mint a hostname per session. Suffixes match on a label boundary, so `.acme.com` never matches `evilacme.com`.

With `Host` validated, **`Origin == Host` is sound again and is restored** — it can only match a page this companion actually served, which is what keeps LAN and reverse-proxied dashboards configuration-free. `isRequestAllowed` owns both checks in the only correct order, so neither caller can run them unsafely, and the `/ws` upgrade now goes through the same gate instead of an origin check alone.

## Deployments — all four keep working, none need manual config

| | Config |
|---|---|
| Docker / localhost | none |
| Another investigator via your LAN IP | none — reads, writes, live updates |
| Railway | none — the entrypoint self-configures from `RAILWAY_PUBLIC_DOMAIN` |
| Killercoda | none — `background.sh` trusts the session hostname by suffix |

`/health` is exempt from the host check for callers sending no `Origin`, so platform health probes (Railway sends `Host: healthcheck.railway.app`) still pass — otherwise every deploy fails. That route returns feature flags only: no case data, no secrets.

`DFIR_ALLOWED_HOSTS` and `DFIR_ALLOWED_HOST_SUFFIXES` join `DFIR_ALLOWED_ORIGINS` on the never-writable-via-API list, so the dashboard cannot widen the gate protecting it.

## Verification

- **5033 tests pass**, `tsc --noEmit` clean.
- Written test-first: each new test was watched failing for the right reason first — the rebound WebSocket upgrade was admitted, the LAN dashboard was refused, the health probe 403'd.
- New coverage: rebinding with and without `Origin`, HTTP and WebSocket; LAN address; configured host, origin-implied host, and suffix; suffix label-boundary evasion; malformed `Host`; the probe exemption and its limits.
- **Not validated with a live exploit**, and the Killercoda suffix (`.killercoda.com,.katacoda.com`) is inferred — the repo does not record the real proxy hostname, so it is worth one tutorial run to confirm.

Docs updated: README env table, `mkdocs-docs/reference/settings.md`, CHANGELOG.


hasamba:
**Killercoda hostname: checked, and the guess is gone.**

The suffix I had used (`.killercoda.com,.katacoda.com`) was inferred — nothing in this repo records the platform's proxy domain. Killercoda's own [network-traffic example](https://github.com/killercoda/scenario-examples/tree/main/network-traffic) shows there is no need to guess: the VM publishes its real public URL at `/etc/killercoda/host` with `PORT` as a placeholder, which is how the platform tells scenarios to build access URLs:

```
sed 's/PORT/80/g' /etc/killercoda/host
```

`killercoda/background.sh` now reads that at start-up and passes the result as both `DFIR_ALLOWED_ORIGINS` and `DFIR_ALLOWED_HOSTS` — the file's format is documented only by example, so passing both is correct whether it yields a full URL or a bare hostname. A missing file logs a warning instead of failing silently.

Suffix matching stays in the guard for platforms that offer no equivalent; the doc and test examples now use a neutral `.lab.example.com` rather than asserting an unverified domain.

The "Killercoda suffix is inferred" caveat in the PR description no longer applies.

Issue #415, opened 2026-08-01
refactor(dashboard): decompose dashboard.html — extract what is standalone, then restructure the shared state


hasamba:
Split out of #384, which had this as one checkbox. It is not a checkbox — it is a state-architecture project, and leaving it inside #384 keeps that issue open indefinitely while the parts that *are* done sit unmerged behind it.

> **Updated 2026-08-05**, after PRs #467–#475. Items 1 and 3 are done; the original framing of item 2 no longer matches the code. The measurements below are re-taken, not copied. The original filing measurement is preserved at the bottom for comparison.

## Where this stands

The binding targets live in `ARCHITECTURE.md:298-300`, not in this body:

| Target | Limit | At filing | Today |
|---|---|---|---|
| `public/dashboard.html` inline JS | 2,000 | 19,203 | **16,634** |
| `public/dashboard.html` inline CSS | 800 | 3,234 | **4 — met** |
| `public/css/dashboard.css` | 800 | did not exist | **3,261** |

Other rows, measured the same way at both ends:

| | At filing | Today |
|---|---|---|
| top-level functions in the inline script | 902 | **724** |
| top-level `const`/`let`/`var` | 408 | **348** |
| modules under `public/js/` (top-level) | 7 | **33** |

2,569 lines have left the inline script. Note the exchange rate: 4,456 lines of new `dashboard-*.js` appeared, so extracted code runs ~1.73× its former size once it carries a header, publication onto `window` and guards. *Lines removed* and *debt retired* are different numbers.

## Item 1 — extract the standalone functions — DONE

Overshot, in fact: 178 top-level function declarations moved, against the ~111 (~816 lines) originally scoped.

The seam is exhausted. 38 functions still match the original "standalone" definition, but only `esc` (`public/dashboard.html:10242`) and `escAttr` (:10246) are transitively movable, and those are pinned inline on purpose — 694 and 263 inline call sites respectively, with `public/js/dashboard-escape.js:32-43` duplicating rather than moving them under a byte-identical drift test. Of the other 36, 17 read page state through a `Dfir*` namespace and 24 delegate to a non-candidate.

## Item 2 — decide who owns dashboard state — OPEN, and reframed

[...]

PR #467, opened 2026-08-04
refactor(dashboard): give the investigation scope window its own owner (#415 tier 2)

hasamba:
First **tier 2** cell of #415. Tier 1 put the case snapshot behind `DfirState`; this puts the
investigation scope window behind a new owner, `DfirScope`.

Scope goes first because it is the one piece of tier-2 state that is not view state at all — it is
server-persisted, with a store, a REST pair and a websocket broadcast behind it. That is also why it
does **not** join the timeline-view owner the rest of tier 2 is headed for.

## The measurement

Resolved through a real lexical scope chain, not by name — three functions declare a *local* `scope`
that has nothing to do with this one (`11524` an HTML string, `13587` an IOC count label, `11732` a
chain label):

| | |
|---|---|
| occurrences of the identifier | 33 |
| → bind to the top-level declaration | **26** |
| → bind to a local shadow | 7 |
| **whole-value writes** | **3** |
| **property mutations** (`scope.start = x`) | **0** |
| reader functions | 3 — `inScope`, `projectScope`, `renderScopeInfo` |

> The census I posted to #415 listed `renderHostRanking` as a fourth writer. It is not — that was the
> shadow at 11524, matched by name. A correction is going on the issue.

Zero property mutations is what makes handing back the value safe, and it is now enforced rather than
measured once.

## Three paths, two commit shapes

The writers differ in what they refresh, so one setter with a redraw flag would change behaviour:

| path | commits | writes the two controls | repaints `#scopeInfo` | redraws |
|---|---|---|---|---|
| `loadScope` | ✓ | ✓ | ✓ | **nothing** |
| `scope_changed` | ✓ | ✓ | ✓ | `render(lastState())` |
| `applyScope` | ✓ | ✗ — it *read* them | ✓ | `/state` refetch → `render`, 20 loaders, super-timeline |

So the module publishes the two shapes they actually differ by — `receive` (the server told us; the
controls are a sink) and `confirm` (the analyst typed it; the controls are the source) — and leaves
the redraw at the call site, because the redraw is exactly what the three disagree about.

**No generic `set`, no `onScopeChange`.** A single subscriber would have to fire the union of all
three. A test pins the published surface so that stays true.

## The projection moves in, and becomes checkable

`projectScope()`/`inScope()` are pure functions of (window, state), so moving them leaves the binding
with no readers outside its owner — the difference between "behind an accessor" and "encapsulated".

The payoff is that they can now be *executed*: `src/analysis/scopeProject.ts` is the same rule written
a second time, the inline copy called itself "the client-side mirror" of it, and nothing checked the
mirror still matched. The parity suite runs the **real server function** — imported, not
re-implemented — against the browser module over six window shapes including inverted and unparseable
bounds. Same correction this issue already made once, for the false-positive rule.

Before committing the extraction I ran a differential test against the original inline function
lifted verbatim by line range: **2400/2400 identical projections**, object identity preserved for the
no-scope case, `contains()` identical on every case.

## The gate

A call-site count cannot express a cell with three legitimate writers, so the rule is not "how many"
but **which** — an allowlist pinning each commit to its owning function.

- `setterRefs()` gains a `namespace` parameter rather than being copied; two gates checking different
  things while both claiming to check writers is how they drift.
- `globalsAddedBy()` gains a `preload` list. This is the first module with a *load-time* dependency,
  so without it the module throws while loading rather than failing to publish.

**All ten new gates were mutation-tested.** One did not fire: the test that `confirm()` leaves the
controls alone pre-set them to the round-trip of the committed window, so a `confirm()` that *did*
write them wrote back the identical string and the test still passed — it was asserting "the value is
right" while claiming to assert "nothing wrote it". It is now a sentinel, and fires.

## Found, not fixed

Behaviour is reproduced exactly. Both are pre-existing and written up on the issue:

- **`applyScope` renders twice.** `broadcastTo` (`src/live/hub.ts:67-80`) iterates every subscriber
  with no sender exclusion, so the POSTing client receives its own echo — which also replaces the
  analyst's typed input with the server's normalised ISO form.
- **`loadScope` redraws nothing** while the websocket branch redraws after identical writes. A
  `/scope` response landing after the first `render()` leaves an unprojected view.

## Gates

`build`, `lint`, `format:check`, `check:size` (18138 → 18108), `check:imports`, `check:boundaries`,
`eval:change-gate`, `check:us-map` all green; route inventory re-baselined (+1 line, the new asset).
Full suite 7336 passed. Four failures are pre-existing and environmental — `sharpVersion` (local
node_modules has sharp 0.33.5/vips 8.15.3, below what the test requires) and one timing-sensitive
ReDoS assertion that **passes when run alone**; neither file is touched by this branch.

### DIFF
--- a/companion/scripts/file-size-ledger.json
+++ b/companion/scripts/file-size-ledger.json
@@ -6,7 +6,7 @@
   "analysis/velociraptorImport.ts": 1503,
   "integrations/velociraptor/velociraptorApi.ts": 1086,
   "public/css/dashboard.css": 3261,
-  "public/dashboard.html#inline-js": 18138,
+  "public/dashboard.html#inline-js": 18108,
   "reports/markdown.ts": 1461,
   "routes/caseLifecycle.ts": 1204,
   "routes/import.ts": 1766,
--- a/companion/src/http/staticAssets.ts
+++ b/companion/src/http/staticAssets.ts
@@ -43,6 +43,10 @@ export const STATIC_ASSETS: Record<string, string> = {
   "/js/dashboard-ioc.js": "application/javascript; charset=utf-8",
   "/js/dashboard-values.js": "application/javascript; charset=utf-8",
   "/js/dashboard-fragments.js": "application/javascript; charset=utf-8",
+  // Tier 2's first owner: the investigation scope window, its projection, and its controls (#415).
+  // Unlike the helpers above this one holds state, so a 404 here is not a missing helper — every
+  // scope read would throw and the dashboard would not render at all.
+  "/js/dashboard-scope.js": "application/javascript; charset=utf-8",
   // The first whole FEATURE module of #415 tier 3, as opposed to the pure-helper modules above.
   "/js/dashboard-tagger.js": "application/javascript; charset=utf-8",
   "/js/dashboard-kev.js": "application/javascript; charset=utf-8",
--- a/companion/tests/architecture/route-inventory.json
+++ b/companion/tests/architecture/route-inventory.json
@@ -478,6 +478,7 @@
     "ROUTE GET /js/dashboard-ioc.js",
     "ROUTE GET /js/dashboard-values.js",
     "ROUTE GET /js/dashboard-fragments.js",
+    "ROUTE GET /js/dashboard-scope.js",
     "ROUTE GET /js/dashboard-tagger.js",
     "ROUTE GET /js/dashboard-kev.js",
     "ROUTE GET /js/a11y/modal-autowire.js",
--- a/companion/tests/dashboard/dashboardApi.ts
+++ b/companion/tests/dashboard/dashboardApi.ts
@@ -272,3 +272,32 @@ export interface StateApi {
     onLastSuperDataChange(fn: (data: Snapshot) => void): () => void;
   };
 }
+
+/** An investigation window. Mirrors the server's ScopeWindow (src/analysis/scope.ts). */
+export interface ScopeWindow {
+  start: string | null;
+  end: string | null;
+}
+
+/**
+ * public/js/dashboard-scope.js — tier 2's first owner (#415).
+ *
+ * Note what is NOT here: there is no `set`, and no `onScopeChange`. The three call sites that write
+ * this window want three different refreshes, so the module publishes the two COMMIT shapes they
+ * actually differ by — `receive` (the server told us; push into the controls) and `confirm` (the
+ * analyst typed it; leave the controls alone) — and leaves the redraw at the call site. A generic
+ * setter plus a subscriber would have to be the union of the three, which is the behaviour change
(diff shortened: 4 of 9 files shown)


hasamba:
Pushed `06be2088` addressing all four findings. Each fix is verified by watching the gate fail — 13 mutations, 13 fired.

**1 (High) — writer allowlist missed `window.DfirScope.confirm(...)`.** Confirmed: the matcher required a bare `Identifier` for the namespace, so it returned zero references for the spelling the browser itself uses.

What made this worth fixing at the root: it is **round three** of the exact failure that `setterRefs`' own comment documents two prior rounds of. Each round patched the case rather than the cause, so each round left a different spelling uncovered. There is now one shared predicate — `denotesNamespace()` — answering "does this expression mean the store", covering the bare name, `window.`/`globalThis.`/`self.`, and the `["Namespace"]` index. `cachedSnapshots()` had the identical hole and takes the same predicate.

That also closes it in the **tier-1** gates, where it had been open since they shipped and nobody had tested it: `window.DfirState.setLastState(null)` was invisible to the single-writer rule. Verified by injecting it — both tier-1 assertions now fail as they should.

Gates fired for: `window.DfirScope.confirm`, `globalThis.DfirScope.receive`, `window["DfirScope"].confirm`, and `const s = window.DfirScope; s.confirm(…)`.

**2 (Medium) — binding check rejected one spelling.** Confirmed: of `let scope = null`, `let scope;`, `const scope = …`, `var scope = {}`, the regex rejected none. A declaration is an AST fact, so `topLevelBindings()` now asks the AST; the moved-function check moved off regex to `functionsOf()` for the same reason. All five spellings now fire, including `let scope = {}, other = 2`.

**3 (Medium) — the promised mutation gate did not exist.** Correct, and the comment claiming it was the worse half of the problem. Your point about the silent no-op is the one that matters: freezing stops the *state* being corrupted but not the *code* being written, and in a non-strict classic script the assignment vanishes with CI green — strictly worse than the throw a strict realm would raise. `getterMutations()` now catches the direct form, the aliased form (`const w = DfirScope.get(); w.start = x`) and the window-rooted computed spelling. The module comment now names the gate instead of promising it.

**Minor — wiring order.** Correct. It now locates the inline block that defines `render()` and asserts the tag precedes it. Worth noting the first attempt was wrong in a way your finding predicted: a single regex spanning the file ran past the first bootstrap block's `</script>` and matched at offset 652 instead of the real script, so it iterates the blocks instead.

On your validation note about local `npm run typecheck` — that matches what I saw, and it is the worktree's `node_modules` symlink (missing Playwright types, plus `sharp` 0.33.5/vips 8.15.3 below what `sharpVersion.test.ts` requires). Exact-head CI with a fresh `npm ci` is the authority there.

Nothing in the four findings touched analyst-facing behaviour, so the extraction itself is unchanged — the differential and parity evidence from the original push still stands.

PR #468, opened 2026-08-04
refactor(dashboard): give the selections and the star cache their own owners (#415 tier 2)

hasamba:
Second **tier 2** step of #415, after [#467](https://github.com/hasamba/DFIR-Companion/pull/467) moved the scope window. Four sets move behind two owners: `DfirSelection` (events / IOCs / findings) and `DfirStarred`.

## Why this one is not a rename

Scope was three whole-value writes and **zero** in-place mutations, so an accessor was a rename. These are the opposite:

| | reassignments | **in-place mutations** | reader fns |
|---|---|---|---|
| `selectedEvents` | 0 | **11** in 8 fns | 21 |
| `selectedIocs` | 0 | **8** in 6 fns | 18 |
| `selectedFindings` | 0 | **7** in 5 fns | 15 |
| `starredEvents` | 2 | **7** in 4 fns | 13 |

Replace-on-write is a real change of mechanism here, so the design had to follow the measurement rather than the plan.

## The loop is why there are bulk operations

Four mutation sites are **inside a loop** — select-all for each of the three panels, and the swimlane rubber band. One commit per iteration is O(n) each, so O(n²) for the gesture. That is not a micro-concern: `tlPageSize` is analyst-selectable and **0 means "show every row"**, so select-all is not bounded by a page, and the swimlane's lane data isn't either. On a real case that is a hung tab, not a slow one.

So the bulk gestures collect first and commit **once** — which is also *fewer* commits than today.

**`addAll`/`removeAll`, not `replace`/`clear`,** and that distinction is load-bearing: select-all ticks the *rendered* rows, so rows selected on another page must survive it. The per-row `.add()` loop did that for free; `replace()` would silently drop them. There is a test, and it fires when `addAll` is degraded to `replace`.

## Two owners, not one

`DfirSelection` is ephemeral view state; `DfirStarred` is a local cache of server-side tags that `loadTags()` re-derives. They look alike, which is exactly the risk — a "clear everything" on the selection owner would otherwise unstar the case.

## No live Set escapes, and freezing is not how

Worth stating because the obvious hardening does nothing: `Object.freeze(new Set())` freezes own properties, and a Set's contents live in internal slots, so a "frozen" Set still accepts `.add()`. The container simply never leaves the closure; `ids()` returns a frozen **array** copy. The three published reads cover every read form the page had — 24 `.has(`, 17 `.size`, 11 `[...set]`.

No refresh in the module, for the same reason as `dashboard-scope.js`: clearing the event selection repaints the timeline and swimlane canvas, clearing the finding selection is pure in-place class toggling that preserves scroll and focus. An owner that redrew on commit would have to pick one.

## The gate

New and specific to replace-on-write: `ownerCalls()` reports every commit and whether a loop encloses it, and **no commit may sit in one**.

`bulkStarIds` is the single allowlisted exception, pinned **by name with its reason** rather than by narrowing the gate to miss it — it awaits one HTTP request per id (`TagsStore.add` is read-modify-write on `tags.json`), so a Set copy is not the cost there, and batching would be actively wrong because each id must land as its own request succeeds. A companion test fails if that allowlist entry ever stops matching anything, so the exemption cannot outlive what it excuses.

**7 mutations, 7 gates fired:** a commit in a loop, a stranded allowlist entry, the old binding returning, `addAll` degraded to `replace`, `ids()` handing back the live Set, the IIFE removed, and a `subscribe()` appearing on the surface.

## Gates

`build`, `lint`, `format:check`, `check:imports`, `check:boundaries` green; route inventory re-baselined (+1 line, the new asset). Full suite **7368 passed**; the 2 failures are the known local `sharp` 0.33.5/vips 8.15.3 in the shared `node_modules`, below what `sharpVersion.test.ts` requires — that file is untouched here and CI installs fresh.

`check:size` is worth a note: it caught this change **growing** the inline script by 15 lines, all of it commentary I had added to the page. The argument moved into the module that owns the rule, which is where it belonged, and the inline script ends up not growing at all.

### DIFF
--- a/companion/src/http/staticAssets.ts
+++ b/companion/src/http/staticAssets.ts
@@ -47,6 +47,10 @@ export const STATIC_ASSETS: Record<string, string> = {
   // Unlike the helpers above this one holds state, so a 404 here is not a missing helper — every
   // scope read would throw and the dashboard would not render at all.
   "/js/dashboard-scope.js": "application/javascript; charset=utf-8",
+  // Tier 2's selection owners: what the analyst has ticked (DfirSelection) and starred
+  // (DfirStarred). Same failure mode as the scope module — these hold state, so a 404 is not a
+  // missing helper, it is every selection read throwing.
+  "/js/dashboard-selection.js": "application/javascript; charset=utf-8",
   // The first whole FEATURE module of #415 tier 3, as opposed to the pure-helper modules above.
   "/js/dashboard-tagger.js": "application/javascript; charset=utf-8",
   "/js/dashboard-kev.js": "application/javascript; charset=utf-8",
--- a/companion/tests/architecture/route-inventory.json
+++ b/companion/tests/architecture/route-inventory.json
@@ -479,6 +479,7 @@
     "ROUTE GET /js/dashboard-values.js",
     "ROUTE GET /js/dashboard-fragments.js",
     "ROUTE GET /js/dashboard-scope.js",
+    "ROUTE GET /js/dashboard-selection.js",
     "ROUTE GET /js/dashboard-tagger.js",
     "ROUTE GET /js/dashboard-kev.js",
     "ROUTE GET /js/a11y/modal-autowire.js",
--- a/companion/tests/dashboard/dashboardApi.ts
+++ b/companion/tests/dashboard/dashboardApi.ts
@@ -273,6 +273,32 @@ export interface StateApi {
   };
 }
 
+/**
+ * One owned set of ids (public/js/dashboard-selection.js).
+ *
+ * Note the absence of anything returning a Set. Every read is a copy or a scalar, which is the
+ * whole point: replace-on-write is only enforceable if the container never leaves the owner.
+ */
+export interface IdSet {
+  has(id: string): boolean;
+  count(): number;
+  /** A frozen COPY, fresh each call. */
+  ids(): readonly string[];
+  /** `on` omitted flips, like classList.toggle. Returns the new size. */
+  toggle(id: string, on?: boolean): number;
+  /** Union in a batch — select-all, the rubber band. ONE commit. */
+  addAll(ids: Iterable<string>): number;
+  removeAll(ids: Iterable<string>): number;
+  replace(ids?: Iterable<string>): number;
+  clear(): number;
+}
+
+export interface SelectionApi {
+  DfirSelection: { events: IdSet; iocs: IdSet; findings: IdSet };
+  /** Starred is a CACHE of server tags, not view state, so it is a separate owner. */
+  DfirStarred: Pick<IdSet, "has" | "count" | "ids" | "toggle" | "replace">;
+}
+
 /** An investigation window. Mirrors the server's ScopeWindow (src/analysis/scope.ts). */
 export interface ScopeWindow {
   start: string | null;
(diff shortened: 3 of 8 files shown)


hasamba:
Pushed `b06de416`. All four findings were correct; each fix is verified by watching the gate fail — **11 mutations, 11 fired**.

**1 (High) — the loop gate missed the shape it was written for.** Confirmed, and this is the uncomfortable one: three of the four sites this migration replaced were `.forEach(cb => …)`, so re-introducing the regression *in its original form* reported `inLoop: false`. The gate could not see the thing it exists to prevent.

The property being tested is "does this run once per element", not "is this a loop keyword". `ITERATION_METHODS` now covers `forEach/map/filter/reduce/some/every/flatMap/find*/sort`, with the callback **arguments** walked at loop depth and the receiver not — so `getIds().forEach(…)` doesn't put `getIds()` in a loop. That also strengthens the tier-1 cached-snapshot gate, which shares `insideLoop`.

Your four sub-points are all closed:
- **Top-level commits** were never scanned at all — `ownerCalls` walked per-function, and a classic script does most of its wiring at top level. It now walks the SourceFile once carrying loop depth. (Proven by breaking the walk and watching the contract test fail.)
- **Aliases and computed members** are now *rejected* by `ownerEscapes()` rather than resolved — the same trade `setterRefs` makes, since following them needs binding-aware analysis and neither has a use here.
- **A wrapper called from a loop** is caught one hop out.
- **A bulk op looping around its own `commit`** is caught against the AST. Worth noting my first version of that check counted `commit(` in the text and *passed* the mutation — one call inside a loop is still one occurrence.

On full transitive reachability: I implemented it, then deliberately bounded it to one hop, and I'd rather state that than have it read as an oversight. Run transitively it reports four things today and none is a per-element cost — `addTag()`/`deleteTag()` reach `deriveStarred()` only through `loadTags()`, which is a fetch (the same reason `bulkStarIds` is allowlisted), and `createNewCase()` reaches `proceedConnect()`, a whole case connect. In a 19,000-line script with names this generic, "can eventually reach" stops predicting "runs per iteration", and a gate whose output is mostly allowlist teaches people to extend the allowlist. The bound is documented at the test so the next reader knows what is *not* covered.

**2 (Medium) — off-page semantic tested only at the API.** Correct. `replace()` had no production caller on a selection and could only ever lose an off-page tick, so I took your first option: it is no longer published there, which makes the regression unrepresentable rather than merely tested. `DfirStarred` keeps it because `deriveStarred()` genuinely rebuilds the whole set. I also did the second option — the three select-all paths are pinned to `addAll`/`removeAll`, and switching one to `clear` now fails.

**3 (Medium) — the binding gate missed valid globals.** Confirmed both: of `let scope = null` / `let scope;` / `const` / `var`, and the block-scoped `var` and bare-assignment cases you named, the helper found none. Both are now found, with the helper's own contract pinned by 8 cases.

**4 (Low) — docs contradicted the code.** Both fixed. I corrected the `dashboard-state.js` tier-2 note *in place* rather than deleting it, because why the plan changed — per-domain refresh costs that a single subscriber cannot express — is exactly what stops the next tier-2 step following the old rule.

One structural change worth flagging: **the analyser now has its own contract tests**. Every hole found in these gates so far — the `async function` shape, the transitive call, the window-rooted namespace, and now the callback loop — was found by a human reading the code and never by a test, because there was no way to ask what the helper says about a given snippet. `scriptFromSource()` makes that possible and 19 cases pin it. That seemed like the actual root cause behind three consecutive audits finding analyser gaps.

No runtime behaviour changed in this round.

PR #469, opened 2026-08-04
refactor(dashboard): give the facet filters an owner and derive the stale prune (#415 tier 2)

hasamba:
First piece of the timeline view, split out and landed on its own. **The split is not arbitrary — this is a prerequisite for the rest**, and that is worth stating before the diff.

## Why this had to come first

Four **renderers** write these sets while drawing them:

```js
for (const s of [...hiddenSources]) if (!allSources.includes(s)) hiddenSources.delete(s);
```

— in `renderSourceFilter`, `renderOriginFilter`, `renderHostFilter` and `renderIocTypeFilter`. A write during a render, inside a loop.

"One coordinated commit per user action" is the shape the whole timeline-view design turns on, and it is **not expressible while the render itself commits**. So this stops here, before the composite actions get written on top of it.

## The measurement

Same shape as the selections in [#468](https://github.com/hasamba/DFIR-Companion/pull/468):

| | reassignments | **in-place mutations** | reader fns |
|---|---|---|---|
| `hiddenSources` | 0 | **7** in 6 fns | 14 |
| `hiddenOrigins` | 0 | **7** in 6 fns | 12 |
| `hiddenHosts` | 0 | **7** in 6 fns | 12 |
| `hiddenIocTypes` | 0 | **6** in 5 fns | 10 |

Same answer: replace-on-write, bulk operations, no live Set out. "hide none" was `facets.forEach(s => hiddenSources.add(s))` — a commit per element, the exact shape #468's gate exists to stop — and is now one `hideAll()`.

## The prune becomes derived — a deliberate behaviour change

The analyst's set stays as they left it; the **effective** set is derived at read time as `hidden ∩ available`.

**The visible difference:** hide a source, re-import a case without it, then import it again. Today the tick is forgotten in between; now it is remembered. That is arguably the better behaviour — they hid it on purpose and never said otherwise — and it is called out here, argued in the module, and pinned by a test named for it rather than smuggled in.

It is also **strictly more correct** in one place: `renderIocTypeFilter` computed `shown = types.length - hiddenIocTypes.size` from the raw size, which was only right immediately after the prune had run. Between renders a stale entry made the count too low. A derived count cannot be stale.

## Two reads, and the difference matters

- `any()` — fast-path guard, lets `renderTimelineEvents` skip a filter pass. A stale entry costs one wasted pass and changes no output.
- `countIn(available)` — the derived intersection, for anything the analyst **sees**: the "3/7" in a button label and whether it is highlighted. A stale entry must never light a filter indicator, so these never use `any()`.

`has` rather than `isHidden` is deliberate: `realSourceCount(sources, hidden)` needs an object with `.has()` and nothing else, so the owner *is* that argument. The one cross-module coupling the census found needs no adapter and no Set escapes.

## Gates

**8 mutations, 8 fired:** a `render*` function committing, the prune re-introduced, a commit in a `forEach`, a wrapper called from a loop, an alias to an owner, `countIn` degraded to the raw size, the old binding returning, and `hideAll` looping around its own commit.

The first of those is the one that matters — it is this step's entire argument, expressed as a check.

`build`, `lint`, `format:check`, `check:imports`, `check:boundaries` green; route inventory re-baselined (+1). Full suite **7422 passed**; the 2 failures are the known local `sharp` 0.33.5/vips 8.15.3 in the shared `node_modules`, in a file this branch does not touch. `check:size`: inline JS **shrank** 18108 → 18104.

## What is left of tier 2

`DfirTimelineView` proper — the composite actions (`revealEvent`, `filterToEventIds`, `setTimeWindow`, `setSearch`, …) over the remaining scalars, `evIdFilter`, and the DOM-backed severity checkboxes. Those are now buildable, because nothing commits mid-render any more.

### DIFF
--- a/companion/scripts/file-size-ledger.json
+++ b/companion/scripts/file-size-ledger.json
@@ -6,7 +6,7 @@
   "analysis/velociraptorImport.ts": 1503,
   "integrations/velociraptor/velociraptorApi.ts": 1086,
   "public/css/dashboard.css": 3261,
-  "public/dashboard.html#inline-js": 18104,
+  "public/dashboard.html#inline-js": 18108,
   "reports/markdown.ts": 1461,
   "routes/caseLifecycle.ts": 1204,
   "routes/import.ts": 1766,
--- a/companion/scripts/file-size-ledger.json
+++ b/companion/scripts/file-size-ledger.json
@@ -6,7 +6,7 @@
   "analysis/velociraptorImport.ts": 1503,
   "integrations/velociraptor/velociraptorApi.ts": 1086,
   "public/css/dashboard.css": 3261,
-  "public/dashboard.html#inline-js": 18108,
+  "public/dashboard.html#inline-js": 18104,
   "reports/markdown.ts": 1461,
   "routes/caseLifecycle.ts": 1204,
   "routes/import.ts": 1766,
--- a/companion/src/http/staticAssets.ts
+++ b/companion/src/http/staticAssets.ts
@@ -51,6 +51,10 @@ export const STATIC_ASSETS: Record<string, string> = {
   // (DfirStarred). Same failure mode as the scope module — these hold state, so a 404 is not a
   // missing helper, it is every selection read throwing.
   "/js/dashboard-selection.js": "application/javascript; charset=utf-8",
+  // The facet filters (source / origin / host / IOC type). Splitting these from the timeline view's
+  // other filters is deliberate: four renderers used to MUTATE them mid-render, and that had to
+  // stop before any "one commit per user action" API could be built.
+  "/js/dashboard-facets.js": "application/javascript; charset=utf-8",
   // The first whole FEATURE module of #415 tier 3, as opposed to the pure-helper modules above.
   "/js/dashboard-tagger.js": "application/javascript; charset=utf-8",
   "/js/dashboard-kev.js": "application/javascript; charset=utf-8",
--- a/companion/tests/architecture/route-inventory.json
+++ b/companion/tests/architecture/route-inventory.json
@@ -480,6 +480,7 @@
     "ROUTE GET /js/dashboard-fragments.js",
     "ROUTE GET /js/dashboard-scope.js",
     "ROUTE GET /js/dashboard-selection.js",
+    "ROUTE GET /js/dashboard-facets.js",
     "ROUTE GET /js/dashboard-tagger.js",
     "ROUTE GET /js/dashboard-kev.js",
     "ROUTE GET /js/a11y/modal-autowire.js",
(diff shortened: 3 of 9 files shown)


hasamba:
Pushed `1507d916`. All five findings were correct — **9 mutations, 9 fired**.

**1 (High) — a remembered filter could hide every row with no control to undo it.** Confirmed, and this is the one that matters: in a forensics tool that is evidence disappearing with no way to get it back, and it came straight from the retention behaviour I argued for.

Worth adding what I found while fixing it: **it is not purely a retention bug.** Hide one of two facets, let the other disappear on re-import, and the same trap springs on code that predates `DfirFacets` entirely — the prune kept the still-present facet hidden and the picker vanished under the `length < 2` threshold. So this is a fix to existing behaviour as well as to the path I introduced.

I took your first option, generalised: **a control the analyst cannot see must not be filtering.** The picker now stays up whenever the effective count is non-zero. That closes both paths with one rule rather than special-casing the sole-facet case.

**2 (High) — the gates failed open.** I reproduced both against the real helpers before touching anything: `renderHosts() { mutate(); }` reported the enclosing function as `mutate`, and `items.forEach(hideOne)` reported `inLoop: false` with `calleesInsideLoops` returning `[]`.

- The renderer check now follows the call graph. Full reachability is right *here* — unlike the loop rule I deliberately bounded on #468 — because the source set is the named `render*` functions and the target is "commits to `DfirFacets`", both specific. No noise.
- `calleesInsideLoops` now collects callbacks passed **by name**; `xs.forEach(hideOne)` has no call expression anywhere inside it, so the identifier *is* the per-element call.
- On "passing the writable owner as an argument": that was a hole **this branch opened itself**, and your finding is what surfaced it — I had `realSourceCount(e.sources, DfirFacets.sources)`, which works and quietly defeats any analysis of who can write. Facets now publish `matcher()`, a frozen `{ has }` view, and the owner is never passed anywhere. That also resolves finding 5 with no cast, since `matcher()` *is* the has-only shape.

**3 (Medium) — `any()` was changing analyst-visible output despite being documented as neutral.** You're right and my documentation was the wrong half of the problem. The same flag drove the timeline's "N of M events" label and the IOC filters-active chip. The saving was one array pass; the cost was the dashboard lying about what the analyst is looking at. **`any()` is removed entirely** — the four `render*Filter` functions return the effective count, which also makes the label agree with the button *by construction* instead of two sites computing it separately.

**4 (Medium) — legacy-binding safeguard incomplete.** Confirmed: it scanned only inline blocks, and the helper missed `function hiddenSources() {}`, `class hiddenSources {}` and `window.hiddenSources = …`. All found now, across every script the page loads.

Your closing observation is the one I've taken most from: *"the added tests do not execute the production renderer interactions, which is why the two visible regressions remain green."* The suite tested the owner's arithmetic and never the renderers that consume it. I've added a check that pins the guard in the page itself, but the general point stands and is the gap I'd want closed before the composite actions land.

Ledger note: the inline script is net-neutral against master (18108 → 18108). My earlier commit on this branch recorded a 18104 shrink; this one's four `return` statements consume it, so I restored the ledger to master's number rather than deleting comments to protect it.

PR #470, opened 2026-08-04
refactor(dashboard): give the timeline view its composite actions (#415 tier 2 complete)

hasamba:
The last tier-2 owner. Ten bindings and 87 read references move behind `DfirTimelineView` — but the move is not the point. **The unit is the user action, not the field.**

## Why not ten setters

Two reasons, both measured rather than asserted:

- **The fields are not the actions.** "Reveal this event" resets six filters, picks a page and repaints. As setters that is six writes whose intermediate states are visible states nobody wants — and the code really did paint them: `filterTimelineToEventIds()` rendered **twice** through its reset before its id filter was even set, so the first two paints showed the filter it had just cleared.
- **The filter is not all in JavaScript.** Severity has *no binding*: it lives in the `.sev-filter` checkboxes and is read from the DOM at render time. A design that models this layer as "these ten variables" cannot be right when part of it is a checkbox.

## One commit, one redraw

`setExcludeTerms` and `applySearch` each called `render()` at their own site **and again** through `renderFalsePositives()`'s tail. One debounced keystroke was two full renders, two `/collection-plan` fetches and four whole-page section sweeps — for the same final state. That collapses. **It is a deliberate behaviour change and the only one in this PR.**

What does *not* collapse is the refresh **set**, which differs per action and is declared per action:

| action | refreshes |
|---|---|
| `setSearch` / `setExcludeTerms` | page · super-timeline · False Positives |
| `setTimeWindow` | timeline · Kill Chain + Phases · super-timeline — **not** the whole page |
| `showOnlyStarred` | timeline only |
| `setCorroboration("timeline")` | timeline; the IOC and finding lenses reach the page |

A single fixed redraw cannot express that, which is why this owner publishes no subscription either.

## The painters are injected — and that is the answer to what three audits kept finding

Every audit of this tier ended on the same note: *the tests do not execute the production renderer interactions, which is why the visible regressions remain green.* That was true, and it was the structural gap rather than any one bug.

This module knows **which** panels an action makes stale and not **how** to paint them; `dashboard.html` registers handlers once via `wire()`. A test registers spies and asserts exactly which panels an action repainted and how many times.

That is what makes the two load-bearing claims checkable at all:

```
it("setSearch repaints the page … ONCE each")            // the collapse
it("has the id filter committed BEFORE the first repaint") // the ordering bug
```

Neither was previously observable outside a browser.

## resetForCase() is not clearFilters()

Connecting to a case deliberately keeps the event-id filter, the severity boxes and the analyst's exclude terms; "clear filters" drops all three. Routing both through one helper would wipe persisted exclude terms on **every** case connect. Preserved, and pinned.

`hydrate()` commits without painting — restoring persisted state runs while the script is still parsing, with no panel to refresh, which is exactly why the original code assigned its bindings directly rather than calling its own setter.

## The read migration was done with the parser

Worth flagging because the first attempt was wrong: a regex rewrite corrupted a localStorage **key** (`"dfir.excludeTerms"`) and an assignment **target**, because a pattern cannot tell an identifier from the inside of a string. Reverted from backup and redone by byte offset over `Identifier` nodes only, skipping write targets and declarations.

## One existing assertion changed

`dashboardHtml.test.ts` pinned `filterFrom = fromIso` inside `zoomToTimeWindow` to show click-to-zoom shares the search bar's path. Both sides now commit through `setTimeWindow`, so the test names **both** rather than implying the link — a stronger assertion than the one it replaces.

## Gates

**8 mutations, 8 fired:** the double render returning, a paint before the id filter is committed, two refresh sets drifting, a case connect wiping exclude terms, `hydrate` painting at bootstrap, a painter left unregistered in the page, and the old binding returning.

`build`, `lint`, `format:check`, `check:imports`, `check:boundaries` green; route inventory re-baselined (+1). Full suite **7462 passed**. Two failures are the known local `sharp` 0.33.5/vips 8.15.3; a third (`anonymize`) passes in isolation — a parallel-load flake in a file this branch does not touch. `check:size`: inline JS **shrank** 18108 → 18097.

**Tier 2 is complete.** Remaining for #415 is tier 3 — the ~231 feature-local bindings, which were blocked on exactly this work.

### DIFF
--- a/companion/scripts/file-size-ledger.json
+++ b/companion/scripts/file-size-ledger.json
@@ -6,7 +6,7 @@
   "analysis/velociraptorImport.ts": 1503,
   "integrations/velociraptor/velociraptorApi.ts": 1086,
   "public/css/dashboard.css": 3261,
-  "public/dashboard.html#inline-js": 18097,
+  "public/dashboard.html#inline-js": 18104,
   "reports/markdown.ts": 1461,
   "routes/caseLifecycle.ts": 1204,
   "routes/import.ts": 1766,
--- a/companion/scripts/file-size-ledger.json
+++ b/companion/scripts/file-size-ledger.json
@@ -6,7 +6,7 @@
   "analysis/velociraptorImport.ts": 1503,
   "integrations/velociraptor/velociraptorApi.ts": 1086,
   "public/css/dashboard.css": 3261,
-  "public/dashboard.html#inline-js": 18108,
+  "public/dashboard.html#inline-js": 18097,
   "reports/markdown.ts": 1461,
   "routes/caseLifecycle.ts": 1204,
   "routes/import.ts": 1766,
--- a/companion/src/http/staticAssets.ts
+++ b/companion/src/http/staticAssets.ts
@@ -55,6 +55,9 @@ export const STATIC_ASSETS: Record<string, string> = {
   // other filters is deliberate: four renderers used to MUTATE them mid-render, and that had to
   // stop before any "one commit per user action" API could be built.
   "/js/dashboard-facets.js": "application/javascript; charset=utf-8",
+  // The timeline view's composite actions — the last tier-2 owner (#415). Unlike the others this
+  // one takes injected painters, so a 404 leaves every filter gesture committing to nothing.
+  "/js/dashboard-timeline-view.js": "application/javascript; charset=utf-8",
   // The first whole FEATURE module of #415 tier 3, as opposed to the pure-helper modules above.
   "/js/dashboard-tagger.js": "application/javascript; charset=utf-8",
   "/js/dashboard-kev.js": "application/javascript; charset=utf-8",
--- a/companion/tests/architecture/route-inventory.json
+++ b/companion/tests/architecture/route-inventory.json
@@ -481,6 +481,7 @@
     "ROUTE GET /js/dashboard-scope.js",
     "ROUTE GET /js/dashboard-selection.js",
     "ROUTE GET /js/dashboard-facets.js",
+    "ROUTE GET /js/dashboard-timeline-view.js",
     "ROUTE GET /js/dashboard-tagger.js",
     "ROUTE GET /js/dashboard-kev.js",
     "ROUTE GET /js/a11y/modal-autowire.js",
--- a/companion/tests/dashboard/dashboardApi.ts
+++ b/companion/tests/dashboard/dashboardApi.ts
@@ -318,6 +318,43 @@ export interface SelectionApi {
 }
 
 /**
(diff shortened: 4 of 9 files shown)


hasamba:
Pushed `041d24cb`. All seven findings were correct — **8 mutations, 8 fired**.

Three of the four High findings are regressions I introduced, and one is a bug I made worse by declaring it intended. Taking them in the order that matters:

**2 (High) — the headline claim did not hold in production.** You're right, and this is the one I'd most want on the record: `renderFalsePositives()` ends with its own `render(lastState())`, the page wired that function directly, so `refresh("all", "falsePositives")` ran two full renders. The PR text claimed the double render was gone while it was still happening.

**Finding 4 is the more important half of that**, and your diagnosis was exact — independent spies removed the very nesting that caused the bug. I've added a suite that models the page's handler nesting, but the assertion that actually closes it is different: **reverting the page alone left those nesting tests green**, because they register their own handlers. That is the same mistake one level out. So there is now a check that dashboard.html *wires* the panel-only form, and I confirmed it fires by reverting the page and watching it go red — the first version of the fix did not.

Every painter spy now also snapshots the state it can see, so moving `refresh()` above the write is caught in **every** action rather than only `filterToEventIds`.

**1 (High) — revealing a group left panels contradicting the controls.** Confirmed. `filterToEventIds` clears filters that scope Findings, IOCs, False Positives and the super-timeline, and refreshed only the controls and the event list. It now refreshes the same set as `clearFilters()`, shared as one list so the two cannot drift apart again. The localStorage half was also real — both clearing paths now drop the persisted copy.

**3 (High) — a case switch kept the previous case's group filter.** You're right that this was an old omission, and right that **newly declaring and testing it as intended is worse than leaving it alone**. An id filter is derived from a case, so carrying it across either blanks the new timeline or shows unrelated evidence under the old label. `resetForCase()` clears it now. The exclude terms and severity boxes genuinely are deliberate — those are the analyst's standing preferences, not the case's — so only the id filter moved.

**5 and 6 (Medium)** both confirmed and fixed: `0` is off throughout, matching the control's own values; and the IOC lens has its own painter, so it no longer refetches the collection plan and resets pagination for a change that only affects the IOC list.

**7 (Medium) — enforcement.** I inverted escape detection as you suggested: an owner may appear as the thing being *called* and nowhere else. Two exclusions were needed to make it usable, and both were found by watching it misfire — a module publishing its own namespace, and the property-*name* half of `window.DfirFacets`, which matches "denotes the namespace" on its own.

**The rest of finding 7 is not fixed**: two-hop loop wrappers, commits via `cell.set()` inside bulk loops, and computed globals still get through. So **"tier 2 complete" was an overstatement about the enforcement and I withdraw it** — the owners are in place, the gates are not yet what that phrase implies.

PR #471, opened 2026-08-04
test(dashboard): close the tier-2 gate gaps withdrawn from #470 (#415)

hasamba:
[#470](https://github.com/hasamba/DFIR-Companion/pull/470) shipped with **"tier 2 complete" withdrawn** — the owners were in place, the gates were not what that phrase implies. This closes the difference.

## I measured first, and the list was not what the review said

Probing the real helpers showed **two of the reported gaps were already closed** by #470's own fixes:

- every detached method-reference form — array, object literal, return value, argument, assignment — reports `passed`;
- `self.X`, and `var` inside `try`/`switch`/label blocks, are all found.

The audit was written against `806b9137`, before those landed. Worth checking rather than assuming, in both directions.

## Three were real, and one worse than reported

**1 — two-hop loop wrappers.** `for (…) outer(x)` where `outer()` calls `inner()` which commits was invisible: the rule looked at direct callees only. Now `reachableWithin()` — **bounded** at two hops, because unbounded reachability over this call graph reports things like `createNewCase()` reaching `proceedConnect()`, where "can eventually reach" has stopped predicting "runs per iteration". The bound is stated at the call site.

**2 — per-element commits inside the modules, and this was worse than reported.** The old check looked for a bare `commit(` inside a hand-listed set of bulk operations, so *both* of these walked past it:

```js
addAll(ids)  { for (const id of ids) cell.set(new Set([id])); }   // member call, not `commit(`
hideAll(ids) { for (const id of ids) put(id); }                   // helper that commits
```

`commitsInsideLoops()` replaces it: a function commits if it calls any commit name bare **or** as a member, closed transitively over the module's own functions, and **every** function is examined rather than a hand-listed few. The list was itself a way to be wrong.

**3 — computed globals.** `globalThis["hiddenSources"] = new Set()` is the same page global the property-access branch already caught, by the one spelling it could not see.

## A deferred callback is not a loop body

This fell out of fixing (1) and is the more interesting half. Two hops reported exactly one offender on the page, and it was noise:

```js
["ncCaseId","ncName","ncInvestigator"].forEach((id) =>
  document.getElementById(id).addEventListener("keydown", () => createNewCase()));
```

That puts a commit three frames inside a `forEach`, but it runs on a **keystroke** — once, later, not once per element. So loop depth now resets when entering a function that is not an iteration callback.

The alternative was an allowlist entry, and a gate whose output is mostly allowlist teaches people to extend the allowlist rather than fix the code. There is a mutation asserting this exclusion did **not** become a blanket escape — a direct commit in a loop is still caught.

## The analyser now has its own contract suite

Six holes have been found in these gates across #415, and **every one was found by a person reading the helper, never by a test** — because the page happened not to contain the shape being missed. `dashboardAstContract.test.ts` keeps each of those bypasses as the regression test it should always have been, each paired with an assertion of what must **not** be reported, since a check that flags everything is as useless as one that flags nothing.

## Gates

**7 mutations against the real page and modules, 7 fired:** two-hop wrappers around a selection and a facet commit, `cell.set()` per element inside `addAll`, a helper committing per element inside `hideAll`, `globalThis["selectedEvents"]` and `window["hiddenSources"]` recreated, and the deferred-callback exclusion not becoming a blanket escape.

`build`, `lint`, `format:check`, `check:size`, `check:imports`, `check:boundaries` green. Full suite **7482 passed**; the 2 failures are the known local `sharp` 0.33.5/vips 8.15.3, in a file this branch does not touch. Tests only — no runtime change.

With this, the enforcement matches the claim, and **tier 2 is complete** in the sense #470 would not assert.

### DIFF
--- a/companion/scripts/file-size-ledger.json
+++ b/companion/scripts/file-size-ledger.json
@@ -6,7 +6,7 @@
   "analysis/velociraptorImport.ts": 1503,
   "integrations/velociraptor/velociraptorApi.ts": 1086,
   "public/css/dashboard.css": 3261,
-  "public/dashboard.html#inline-js": 18104,
+  "public/dashboard.html#inline-js": 18102,
   "reports/markdown.ts": 1461,
   "routes/caseLifecycle.ts": 1204,
   "routes/import.ts": 1766,
--- a/companion/tests/dashboard/dashboardAstContract.test.ts
+++ b/companion/tests/dashboard/dashboardAstContract.test.ts
@@ -14,10 +14,11 @@
 import { describe, expect, it } from "vitest";
 import {
   buildCallGraph,
+  ownerEscapes,
   calleesInsideLoops,
   commitsInsideLoops,
   ownerCalls,
-  reachableWithin,
+  reachableFrom,
   scriptFromSource,
   topLevelBindings,
 } from "../helpers/dashboardAst.js";
@@ -25,7 +26,7 @@ import {
 const COMMITS = ["toggle", "addAll", "removeAll", "clear", "showAll", "hideAll"];
 
 describe("the loop rule follows a wrapper, but not a deferred callback", () => {
-  it("catches a commit two hops from a loop callee", () => {
+  it("catches a commit several hops from a loop callee", () => {
     const s = scriptFromSource(
       "p.js",
       `
@@ -37,7 +38,7 @@ describe("the loop rule follows a wrapper, but not a deferred callback", () => {
     const graph = buildCallGraph([s]);
     let caught = false;
     for (const callee of calleesInsideLoops([s])) {
-      const reach = new Set([callee, ...reachableWithin(graph, [callee], 2)]);
+      const reach = new Set([callee, ...reachableFrom(graph, [callee])]);
       for (const c of committers) if (reach.has(c)) caught = true;
     }
     expect(caught, "two-hop wrapper still invisible").toBe(true);
@@ -54,7 +55,7 @@ describe("the loop rule follows a wrapper, but not a deferred callback", () => {
     const graph = buildCallGraph([s]);
     const hits: string[] = [];
     for (const callee of calleesInsideLoops([s])) {
-      const reach = new Set([callee, ...reachableWithin(graph, [callee], 2)]);
+      const reach = new Set([callee, ...reachableFrom(graph, [callee])]);
       for (const c of committers) if (reach.has(c)) hits.push(`${callee}->${c}`);
     }
     expect(hits, "a keystroke handler is not a per-element commit").toEqual([]);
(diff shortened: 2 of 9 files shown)


hasamba:
Pushed `23bfcee1`. All six enforcement findings plus the runtime defect — **10 mutations, 10 fired**.

**The runtime defect first, because it is merged and analyst-facing.** You're right, and it's mine: `setSearch()` trims and lower-cases, and #470's painter wrote that committed value back into the live input 300ms later. `"foo "` + pause + `"bar"` really does produce `foobar`. The original `applySearch()` read the input and never wrote it — I added the write when building the painter.

The painter now touches only the clear button. Emptying the box is a *reset* (what `resetTimelineViewFilters()` did), so it is its own painter that only the clearing actions ask for. Gated, and I confirmed the gate fires by restoring the bad write.

**1 (High) — and it was a regression this PR introduced.** `commitsInsideLoops()` skipped every nested function, so `addAll(ids) { ids.forEach(id => cell.set(…)) }` returned no offender — the exact per-element commit the owners exist to prevent, and one the check it *replaced* caught. That's the worst kind of finding and you were right to block on it.

**2 (High) — the deferred-callback exception was far too broad**, and your probes named the cases exactly. My reasoning was wrong at the root: I treated *deferral* as the property that matters, when it is *being driven by something other than the loop*. `queueMicrotask` defers and still runs N times. Only listener registration and `el.onclick =` qualify now, so IIFEs, parenthesized callbacks, `setTimeout`, `queueMicrotask` and `.then` all count again — and the mirror case, a named handler registered in a loop, is no longer reported falsely.

**3 (High)** — confirmed; selection and scope scanned only the inline blocks while the other two suites scanned everything. Fixed, so the contract suite's claim is now true of every suite rather than most of them.

**4 (Medium) — the bound is gone entirely, not raised.** This is the part I'd most want you to check. Two hops was a documented fail-open that existed *only* because unbounded reachability reported noise — and the noise turned out to be the modelling error in (2), not a fact about the code. With handlers no longer counted as per-element, unbounded reachability reports **nothing** on this page. So the bound was limiting what could be caught and nothing else. `reachableWithin()` is deleted: with the bound gone it had no caller.

**5 and 6 (Medium)** — all four binding forms and both owner-evasion forms confirmed and closed. `.call`/`.apply`/`.bind` are transparent now, and only the namespace's *own* definition is exempt from the escape rule, so assigning a replacement to a member is reported.

One process note: while fixing the size ratchet I briefly set the ledger to 18108, which would have **raised** it above master's 18104 — precisely what that ratchet forbids. I folded the two painters onto one line each instead (this file's existing idiom), so the inline script ends at 18102, a genuine shrink.

PR #472, opened 2026-08-04
refactor(dashboard): move eight whole features out of the inline script (#415 tier 3)

hasamba:
**725 lines leave `dashboard.html`** — the first tier-3 step, and the one tiers 1 and 2 existed to unblock. `#inline-js`: 18102 → 17377.

## The measurement changed the plan

Clustering the movable functions by call graph produces **one 128-function component of ~2,307 lines with 282 inbound calls** — the render layer connected to itself, not a feature. So "move the functions that touch nothing" is the wrong unit.

The ADR already had the right one: a feature's bindings travel **with** it, so a function reading its *own* feature's state isn't blocked at all. What blocks a feature is a binding shared with something else. Measured that way, [nine features have zero escaping reads](https://github.com/hasamba/DFIR-Companion/issues/415#issuecomment-5181227159).

| module | fns | own state |
|---|---|---|
| `dashboard-anomalies.js` | 4 | cached response, debounce timer |
| `dashboard-sessions.js` | 4 | segmentation, timer, ephemeral summaries |
| `dashboard-compliance.js` | 8 | cached response, timer |
| `dashboard-d3fend.js` | 3 | cached response, timer |
| `dashboard-geo.js` | 8 | live Leaflet map, layers, tile URL, init flag |
| `dashboard-custody.js` | 4 | records, failed paths, verified-at |
| `dashboard-backup.js` | 2 | — |
| `dashboard-collection-plan.js` | 2 | — |

**The ninth is held back.** Ticket integrations has four `const overlay = document.getElementById(...)` bindings captured at load, and these are `<head>` scripts — those captures need making lazy first. That's a change to how the feature initialises rather than a move, so it gets its own diff.

## An IIFE, unlike the tagger and KEV modules

Those hold no state, so top-level declarations were harmless. These own state, and a top-level `let` in a **classic** script joins the shared global lexical environment — reachable by name from every other script, the exact hazard `dashboard-state.js` sets out. Wrapping them is what makes "feature-local" a fact rather than an intention.

## Three things the census got wrong

All found by checking rather than trusting it, and each would have shipped a bug:

- **`markAnomalySpikeFalsePositive` and `summarizeSession` *are* called from outside** — from listeners. The census counts calls between top-level functions, so a listener registration is invisible to it. Publishing its list would have left two dead controls.
- **`sessionsCollapsed`** sits inside the sessions block, but the timeline header's collapse-all control reads it. Shared state — it stays in the page. Taking it would have broken that control silently.
- **`initCustodyButtons` is a named IIFE**, not a declaration, so the walk over top-level declarations never saw it — and it writes two of custody's three cells.

## A real timing bug the move would have introduced

`initCustodyButtons` ran at its old position in the inline script, **after** the custody markup. Auto-running it from a `<head>` module would query for buttons that don't exist yet and wire nothing — no error, just a feature that silently stops working. It's a plain function now, called from the page where the IIFE used to run. Caught by the vm loader, then verified as a real browser-ordering problem rather than a test artifact.

## Two helper fixes this forced

`dashboardClientSource()` now reads the tagged scripts **from the markup** instead of a hard-coded list. That list named the original eight helpers and none of the tier-2 owners or these modules, so four suites asserting "the dashboard does X" broke the moment X moved into a file it didn't mention — the same hole `dashboardScripts()` closed, for the same reason. The nonce check deliberately still reads `dashboard.html` alone: it's about markup, and scanning the whole client source finds `<script` inside these modules' own comments.

`functionsOf()` gains `declaration`, because the ACTIONS dispatch table is full of `setComplianceDiscovered: (el) => setComplianceDiscovered(el)` entries whose arrow takes the property's name. Those must **stay** when the function they call moves out; without the distinction they read as leftovers.

## Gates

**7 mutations, 7 fired:** a published name not published, state leaking out of the closure, the closure removed, half a feature left behind, a binding stranded in the page, custody back to an auto-running IIFE, and `sessionsCollapsed` dragged along.

`build`, `lint`, `format:check`, `check:size`, `check:imports`, `check:boundaries` green; route inventory re-baselined (+8). Full suite **7546 passed**; the 2 failures are the known local `sharp` 0.33.5/vips 8.15.3 in a file this branch doesn't touch.


hasamba:
Pushed `9e88ba67`. All four findings were correct — **5 mutations, 5 fired**.

**1 (High) — the isolation test certified nothing, and you're right that this was the primary guarantee.** I wrote text assertions for these modules when the scope, selection, facets and timeline-view suites all *load* theirs in the sandbox helper. The `})();` → `});` case is the one that stings: a valid, never-invoked function expression that publishes not one global, with every assertion still green.

The gate now asserts the **exact** set of globals each module adds (which catches the never-invoked case, the `window.debugFeatureState` leak, and the missing `geoTileUrl` in one check rather than three), calls each published name, and asks the *running* module whether private names are reachable.

Your fourth bypass needed a fifth check, and it's the interesting one: **deleting a private declaration is invisible to every load-time assertion**. The name only becomes an implicit global when the function first *assigns* to it, so the global set is unchanged, a bare reference still throws, and the module starts leaking only once an analyst uses it. There's now a check that each private name is genuinely declared. I confirmed it fires; the other four checks did not.

Stranded-binding search now covers every loaded script rather than the page alone.

**2 (Medium) — my comment was simply false.** `initCustodyButtons()` and `verifyCustodyOnOpen()` are called unguarded and `loadCaseBackups` is registered by name, so a 404 is a `ReferenceError` that takes out the rest of that init or case-connect flow. I took your second option — these are as critical as the helpers, and the comment says so now. Guarding just these three while twenty other modules are equally unguarded would have been inconsistent rather than safer.

**3 (Medium) — both parts correct.** The helper claimed "every first-party script" and reads only the `dashboard-*` tags; that claim is gone. And your `readAsDataURL` example is exact — `dashboard-values.js` contains the same call, so the assertion would have survived deleting the logo's own encoder. Those two now read `dashboard.html` directly, and the helper documents that a concatenation can't answer a location-sensitive question.

**4 (Low) — acknowledged, not changed.** These have to be parser-blocking: the inline script calls their names by bare name while parsing, which is what rules out `defer`. Same trade every module in this issue makes; I've noted it in the commit rather than leaving it implicit.

On the inherited #471 items — those are real and I'm not going to claim otherwise in this PR, which changes no analyser code. They belong in a follow-up against the analyser rather than being folded into a tier-3 extraction.

Thanks for loading it in real Chromium; that's the check I can't run here, and it's the one that would have caught a broken extraction fastest.

Issue #482, opened 2026-08-05
test(dashboard): check:imports cannot see the 26 extracted modules — they use window globals, not imports


hasamba:
`npm run check:imports` includes `public/js/**` specifically so that "the first cycle between extracted feature modules fails a PR" — stated in `ARCHITECTURE.md:235-236` and again in `companion/scripts/check-imports.mjs` at the `PUBLIC_JS` constant. For the 26 modules #415 has actually produced, that gate can never fail.

### Measured

- **0 of 26** `public/js/dashboard-*.js` files contain any `import` or `export` statement.
- **0 of their 26 `<script>` tags** carry `type="module"` — they are classic scripts.
- `check-imports.mjs:25` resolves the graph "by regex over the import statements", so it sees **26 nodes and 0 edges**.

The real dependency graph between those modules is roughly **35 edges**, carried by `window` globals rather than imports — e.g. `dashboard-swimlane.js` depends on escape, selection, ioc, state, time, values and timeline-view; `dashboard-compliance.js` on escape, ioc and fragments. There are 0 cycles today, so nothing is currently hidden.

### Why it was written this way

The gate's own comment records the assumption:

> They import NOTHING from each other today, so this root contributes zero cycles — which is exactly why it is wired in now. As #384 pulls features out of dashboard.html's 19k lines of inline script, these modules start importing one another, and the first cycle should fail a PR rather than be discovered later from a blank page.

The extraction pattern that #415 actually settled on is different: classic scripts publishing onto `window`, so a feature survives a sibling 404 (see `public/js/dashboard-facade.js`). That is a deliberate and defensible choice — but it routes every inter-module dependency around the gate.

### Why it matters now

#415's remaining work is roughly 30 more feature extractions on this same pattern (14,634 inline JS lines to go against `ARCHITECTURE.md:298`'s 2,000-line target, at a measured ~489 lines/PR). Those modules join a dependency graph nothing governs, and the failure mode the gate was built to prevent — "discovered later from a blank page" — is exactly what PR #475 hit twice.

### Done looks like

Either:

1. **Extend the gate** to build the `public/js` graph from cross-module global references instead of import statements. The repo already has the machinery: `moduleGlobals()` in `companion/tests/helpers/dashboardAst.ts` maps every module to the names it publishes, and `buildCallGraph()`/`reachableFrom()` at :1753/:1766 do the traversal. The 35 edges are computable today.
[...]

Issue #480, opened 2026-08-05
test(dashboard): domAccessOutsideFunctions() has three blind spots and runs against one feature of ten


hasamba:
Two structural checks in the dashboard suite are still bypassable.

### The declaration census is regex-based

`companion/tests/dashboard/dashboardFeatureModules.test.ts` matches function declarations by regex. Putting a comment between `function` and its name hides a duplicate declaration:

```js
function /* hi */ initTicketIntegrations() { }
```

All relevant tests pass with the duplicate present. This is the same class of failure PR #473 already fixed twice elsewhere — a check on prose rather than on structure — and the suite has a TypeScript AST available to it (`functionsOf()` in `tests/helpers/dashboardAst.ts`) that answers the question properly.

### `domAccessOutsideFunctions()` has gaps and is barely applied

`companion/tests/helpers/dashboardAst.ts:1340` misses:

- synchronous callbacks
- aliased references
- parameter propagation
- parser-time microtasks

and its production assertion is applied only to the ticket integrations, not to the other feature modules that have the same shape.

### Done looks like

The census asks the AST, not a regex, with a contract test using the comment-injection mutation above. `domAccessOutsideFunctions()` covers the four shapes listed, each with a mutation proving it fires, and the assertion is applied to every feature module rather than one.

Found by an independent audit of PR #475 at `cfdef75e`. Sibling issues: #476, #477, #478, #479.

hasamba:
**The declaration-census half of this issue landed in #484 (`f7a1fec7`).** Only the `domAccessOutsideFunctions()` half remains open.

What was done:

- The census now asks the parser, not a regex. Verified on the real module: injecting a comment between the keyword and the name of `dashboard-tickets.js`'s `initTicketIntegrations` dropped the regex census from 15 declarations to 14 and lost that name; the AST census kept it.
- Review of that PR found the census was wrong in a second, larger way that this issue did not mention: **both** sides counted `function f(){}` declarations only, so `const loadAnomalies = () => {}` restored in the inline script shadowed the module's published function at every inline call site with the whole suite green. `functionBindingsOf()` now counts declarations *and* function-valued variable bindings, while still leaving the ACTIONS dispatch table alone — a property is not a binding.
- The seam itself (`duplicateBindings()`) is now driven by synthetic module + inline sources, because asking it about the real modules returns `[]` whether the gate works or not.

[...]

Issue #479, opened 2026-08-05
test(dashboard): the Swimlane can be fully unwired with its suite green — no fixture runs its initializer


hasamba:
`dashboard-swimlane.js` has a `FEATURES` manifest entry recording `initSwimlane` as its initializer, but no test ever *runs* that initializer against a DOM fixture. The ticket integrations have exactly this (`describe("initTicketIntegrations")` loads a fixture, calls the initializer, and names every control it wires); the swimlane does not.

### Two mutations, both green

1. Delete the guarded initializer call at `public/dashboard.html:12150` — the feature never wires, all 101 feature tests pass.
2. Delete the `swimlaneZoomIn` listener at `public/js/dashboard-swimlane.js:437` — a control is dead, all 101 feature tests pass.

### Why this one matters more than it looks

The swimlane is the 461-line feature extracted out of the inline script in PR #475. The extraction is exactly when a wiring regression is most likely, and it is the one feature with no execution coverage.

### Done looks like

A DOM fixture that executes `initSwimlane()` and asserts, by name, every control and listener the feature owns — the file registers **17** listeners, so a `length > N` threshold is not enough. Per PR #475's own lesson: a threshold cannot tell "all seventeen" from "any five", and reads a duplicate registration as more evidence of success.

The fixture must also cover the `querySelectorAll` path if the feature uses one; a fixture answering `[]` for everything silently certifies unwired controls.

Found by an independent audit of PR #475 at `cfdef75e`.

hasamba:
Reopened — GitHub closed this automatically when #486 merged, because that PR's body read "Closes #479's **first mutation**" and the linkifier only saw "Closes #479". Half of this issue is done; half is not.

**Done in #486 (`aef00032`):** the first mutation — deleting the guarded `initSwimlane()` call from the page. That previously passed all 101 tests in the feature suite. It now fires `dashboard-swimlane.js: the page calls initSwimlane exactly once, at load`, which asks the AST for a call on a path that actually runs at load, rather than `callsByName()`'s "reachable from somewhere".

Also landed there: `dashboard-custody.js` had no `initializer` field at all, so every initializer check silently skipped it; and the manifest can no longer omit a whole feature, because the expected set is derived from the page's `<script>` tags.

[...]

Issue #477, opened 2026-08-05
test(dashboard): callsByName() treats any top-level mention as reachability, so an uncalled initializer passes


hasamba:
`callsByName()` in `companion/tests/helpers/dashboardAst.ts:1282` answers "is this name mentioned at the top level" and is used to answer "is this initializer actually reached". Those are different questions.

### Reproducing

This passes every relevant test while `initTicketIntegrations` is never called:

```js
function dead() {
  initTicketIntegrations();
}
void dead;
```

The mention inside `dead()` satisfies the check; nothing invokes `dead`.

### Why it matters

This is the check standing between the repo and an initializer that silently stops running. A feature can be fully unwired — no listeners, no fetches, no published entry point — with the suite green.

### Done looks like

Real call and registration edges: a `CallExpression` whose callee resolves to the name, or a registration that will invoke it (`addEventListener("...", name)`, `el.onclick = name`), not an arbitrary identifier reference. Reuse the reachability from the `walkLoadTime()` work if that lands first.

Mutation tests required, including the `void dead;` shape above.

Found by an independent audit of PR #475 at `cfdef75e`.

PR #488, opened 2026-08-05
refactor(dashboard): 103 extractions, 7,666 → 2,826 lines — needs a scope decision on the 2,000 target (#415)

hasamba:
> **Status: needs a scope decision before it can be called done.**
>
> `dashboard.html#inline-js` is **7,666 → 2,826** (−63%) across 103 extractions. The issue's target is 2,000.
>
> The remaining 826 lines are the page's spine, and each item has a specific reason it cannot move under the tier-3 recipe:
>
> | remaining | why it stays |
> |---|---|
> | `render`, `connect`/`proceedConnect` | called from the state path, the websocket handler and a dozen features |
> | `renderIocs`, `renderTimelineEvents` | the two render entry points every feature draws through |
> | `sortTimelineEvents` | a guard that skips a SORT shows a timeline in the wrong order with no indication — evidence misrepresented, not a decoration dropped |
> | page state + shared vocabulary | `ws`, `SEV`, `esc` (125 call sites), the pagination cursors |
> | the click router | routes to every feature; stub it and nothing on the page is clickable |
> | ~75 guard stanzas | one per extraction; collapsing them blinds the lifecycle gate that caught two silent breakages |
>
> **Two things to decide:**
> 1. **Is 2,000 the right number?** It counts the spine identically to features. With the spine inline, the honest target is ~2,800.
> 2. **If it must be 2,000**, that is a spine modularisation — a different change, needing a decision about what a missing module may cost when the answer is "the page does not render".
>
> I twice declared a floor in this work and was wrong both times, finding ~1,500 more lines by re-measuring. The difference now is that every remaining item has a named reason rather than a category.

`dashboard.html#inline-js` **16,634 → 6,281**. Eighty-two features extracted, three gates added, and an inventory of what is actually left — because the way features were being chosen turned out to be the weakest part of the process, and the last two extractions were the first ones picked from measurement rather than by eye.

## The eighty-two extractions

| module | lines out | shape |
|---|---|---|
| `dashboard-mcp.js` | 318 | 13 wiring calls → `initMcp()` |
| `dashboard-hunt-profile.js` | ~120 | plain move |
| `dashboard-derived-panels.js` | ~240 | four panels, one module |
| `dashboard-deep-pass.js` | 174 | two state escapes **closed**, not published |
| `dashboard-super-timeline.js` | 591 | largest move; `initSuperTimeline()` |
| `dashboard-playbook.js` | 380 | 4 wiring calls → `initPlaybook()` |
| `dashboard-diagnostics.js` | 327 | first picked from the inventory; `initDiagnostics()` |
| `dashboard-report-versions.js` | 216 | `initReportVersions()` |
| `dashboard-settings-tools.js` | 253 | three Settings panels under one banner |
| `dashboard-sigma-hunt.js` | 325 | `showToast` deliberately left behind |
| `dashboard-push-token.js` | 47 | banner covered two features; only 47 of 222 lines moved |
| `dashboard-analysis-runs.js` | 156 | first picked with the cohesion check |
| `dashboard-report-templates.js` | 151 | two controls bound by value |
| `dashboard-adversary-hints.js` | 129 | no initializer — measured, not assumed |
| `dashboard-gap-hypotheses.js` | 117 | no initializer |
| `dashboard-narrative.js` | 161 | two adjacent bindings deliberately left behind |
| `dashboard-host-ranking.js` | 115 | one delegated listener |
| `dashboard-nsrl.js` | 98 | nine controls, six bound by value |
| `dashboard-views-editor.js` | 109 | not the same thing as "view presets (#142)" |
| `dashboard-import-undo.js` | 52 | collects two bindings a previous extraction left owed |
| `dashboard-case-lifecycle.js` | 149 | 113 of them were a self-calling wiring IIFE |
| `dashboard-enrichment.js` | 77 | a binding that was also an element id |
| `dashboard-unified-import.js` | 154 | wiring only — the initializer is the module |
| `dashboard-asset-overrides.js` | 111 | wiring only |
| `dashboard-encrypted-export.js` | 57 | modal, bound from the shared block |
| `dashboard-redacted-export.js` | 52 | same shape |
| `dashboard-explain-event.js` | 60 | no initializer |
| `dashboard-timeline-gaps.js` | 51 | no initializer |
| `dashboard-unified-export.js` | 76 | wiring only |
| `dashboard-manual-add.js` | 49 | one helper, eight wiring statements |
| `dashboard-correlation-profile.js` | 48 | a window table that was a global |
| `dashboard-settings-modal.js` | 78 | a localStorage key that was a global |
| `dashboard-toolbar-responsive.js` | 43 | wiring only |
| `dashboard-save-template.js` | 38 | bound by value from the shared block |
| `dashboard-zip-archive.js` | 35 | sibling of the encrypted export |
| `dashboard-timeline-display.js` | 38 | the row toggle checkboxes |
| `dashboard-wizard-ai-step.js` | 88 | fourth self-calling wiring IIFE |
| `dashboard-disk-warning.js` | 32 | a `const` lookup that had to move with its wiring |
| `dashboard-ioc-whitelist.js` | 47 | five of six bindings; the sixth stayed |
| `dashboard-ioc-blocklist.js` | 23 | export modal |
| `dashboard-case-template-picker.js` | 27 | bound by value |
| `dashboard-search-scope.js` | 224 | largest wiring-only block; two lines rehomed |
| `dashboard-import-case.js` | 104 | all of it is the initializer |
| `dashboard-exec-summary.js` | 61 | no load-time code of its own |
| `dashboard-import-severity.js` | 60 | same |
| `dashboard-merge-picker.js` | 52 | |
| `dashboard-preflight-banner.js` | 52 | |
| `dashboard-data-act.js` | 133 | the CSP dispatch layer |
| `dashboard-tooltip.js` | 34 | |
| `dashboard-presentation-mode.js` | 22 | 4 of its 5 load-time statements were other features' |
| `dashboard-finding-workflow.js` | 59 | one banner **split** into two features |
| `dashboard-pinned-findings.js` | 152 | the other half |
| `dashboard-palette-registry.js` | 111 | second banner **split** |
| `dashboard-section-order.js` | 82 | the other half |
| `dashboard-query-translator.js` | 93 | one banner **split three ways** |
| `dashboard-scope-apply.js` | 33 | the singleton cluster |
| `dashboard-ai-toggle.js` | 43 | the third |
| `dashboard-velo-monitors.js` | 146 | banner split; four bindings **rehomed** |
| `dashboard-velo-bundles.js` | 173 | the other half |
| `dashboard-import-progress.js` | 49 | holds the read-only import guard |
| `dashboard-ai-status.js` | 66 | the singleton that shared its banner |
| `dashboard-custom-importers.js` | 90 | `wlApplyToCase` went home from here |
| `dashboard-view-presets.js` | 177 | not the views *editor* of the same issue |
| `dashboard-env-settings.js` | 207 | owns the snapshot that decides what gets rebuilt |
| `dashboard-memory-next-steps.js` | 94 | no load-time code of its own |
| `dashboard-delete-case.js` | 84 | I got its boundary wrong; the gate said so |
| `dashboard-case-password.js` | 71 | publishes only its initializer |
| `dashboard-setup-wizard.js` | 255 | two consts in, two functions out |
| `dashboard-bulk-select.js` | 81 | UI over the tier-2 selection store |
| `dashboard-bulk-ioc.js` | 39 | its sibling |
| `dashboard-case-unlock.js` | 61 | the page's startup sequence stayed |
| `dashboard-new-case.js` | 97 | out of a 692-line banner, 595 of it not this feature |
| `dashboard-case-templates.js` | 79 | **closes** a cross-module state escape |
| `dashboard-velo-triage.js` | 464 | **closes seven**, three ways |
| `dashboard-keyboard-nav.js` | 134 | unblocked by moving two tables to their readers |

Each one follows #414's recipe: classic script publishing onto `window`, registered in `STATIC_ASSETS`, route inventory re-baselined. Each was verified in Chromium twice — with the module present, and with it 404ing — because CI was green on all three of the page-killing bugs that pass caught earlier in this series.

Note the exchange rate the issue warns about: 3,011 lines left the inline block and the ten new modules total 4,539 lines including the facade's growth. Extracted code runs larger once it carries a header, `window` publication, guards, and prettier's formatting (the inline block is not prettier-formatted; the modules are).

## `initPlaybook` is deliberately not stubbed

The facade stubs a missing module's work functions so one 404 cannot end a refresh chain. It must never stub an **initializer**: that is the module's own sentinel, and a stub would swallow the "Playbook unavailable" chip that is the only way a missing file announces itself. A stub may replace work; never evidence.

## The inventory, and why it changed the plan

#415 says the first task is an inventory of the remaining features, "not an extraction". Six features had been picked by eye and measured by hand instead. `npm run inventory:dashboard` is that inventory: **94 banner-delimited sections covering all 14,541 lines** of the large inline block, each with its size, function and binding counts, and the three measurements that decide whether it can move.

**State escapes are not escapes.** A function called from outside its block is the ordinary case — the module publishes it onto `window` and no call site changes. A mutable binding read from outside is the blocker. Counted together they mis-rank the list badly. This PR's own earlier description put `sec-playbook` at "12 escaping" by the hand method; measured properly it had **zero**, and it extracted cleanly.

**Measured on the AST, not the text.** Both blocks measured by hand before this existed had apparent escapes that were a mention inside a comment and an HTML `id` attribute. Grep reports clean blocks as blocked.

**It sees the already-extracted modules too.** A module in `public/js` calling a page function by bare name is not an identifier anywhere in the inline AST, so without a cross-file pass the name reads as block-local and the extraction silently fails to publish it — in a browser only. The first version of that scan resolved the wrong directory, found nothing anywhere, and looked correct because the headline did not move; two contract tests pin it now. Fixed, it revealed real state escapes (`ST_PAGE`, `_jobsCache`) and two blocks holding shared machinery.

**It predicts which blocks need an initializer.** `moduleScopeDom: 0` means a block wires nothing — not that nothing wires *it*. The page binds every modal's controls from one shared block hundreds of lines away, so moving a feature's functions out leaves bare references evaluated at load. Both of the last two extractions hit this. `boundElsewhere` now catches it: **43 of the ready sections need an initializer**, so it is the common case, not a quirk. Three extractions in a row since have confirmed it.

**The split itself is a tool now.** `npm run split:dashboard-section` decides what becomes the module body and what becomes the initializer, by parsing. I got that split wrong by hand on the Settings block — a search for lines beginning `document.` found three of six load-time statements, and the three it missed were bare `{ … }` blocks wiring a button each, which would have shipped three permanently dead controls. The rule is "is this a declaration", not "does it touch the DOM", so a timer or a fetch kicked off at load is caught too. It also refuses a line range that cuts a statement in half — line numbers go stale on every extraction, and it caught exactly that on first use. And it warns about `const el = document.getElementById(…)` at module scope: a VariableStatement that reads as module body and, in a `<head>` script, evaluates to null before the markup exists.

**Fan-out catches bad boundaries.** Spot-checking the inventory's own headline found `render()` — the page's central render function, 22 call sites across the file — sitting inside the "Now investigator cockpit (#375)" banner. By state escapes alone that section read as 437 ready lines; extracting it as written would have moved `render()` into a feature module. Banner comments are the author's grouping, not a guarantee of cohesion, so each published function now carries its external reference count and "ready" means both halves.

Result: **44 sections / 4,829 lines are ready today** — a very different picture from the issue's "roughly 30 more PRs" framing, and a concrete ordered work list rather than an estimate.

## Gates added

- **Facade coverage** — every feature name in a load-time refresh fan-out must be stubbed. Took three attempts; the first two passed their own mutation test (#476's blind spot) and produced 41 false positives from tier-1 helpers respectively.
- **Cross-file `@keyframes`** — animations resolve by name across the whole CSSOM, so the CSS split needed a test that a keyframe defined in one file still reaches its user in another.
- **Name resolution** — 2,570 identifiers reduced to 195 free, then 50 unaccounted, 48 of them built-ins. The page is clean. This one came from a review question: why not just test that every function the page calls exists?
- **Inventory coverage** — the inventory must account for every line, verified by running it against a dashboard built to violate that and confirmed by deleting the guard and watching the test fail.

## CI status

GitHub has been in a Partial System Outage across this branch's last several commits. The most recent head that CI actually ran — `f57e3a7e` — passed **Companion lint + format + structure** and **Companion build + test**; only the Extension job was cancelled at runner allocation. Since then GitHub has dispatched **no workflow runs at all** for new pushes (zero check-runs on the head commit).

**Everything after `f57e3a7e` is verified locally only** — 8,597 tests / 570 files plus lint, format:check, typecheck, check:size and build. It needs a CI re-run before merge.

## Verification

Full suite **8,610 tests / 570 files**. `typecheck`, `format:check`, `check:size` (ledger updated for the shrink), `check:imports`, `check:boundaries`, `lint` all clean. Route inventory re-baselined for each new asset.

One legacy assertion had to stop pinning `t =>`: prettier formats extracted modules and does not format the inline block, so the same code reads `t =>` in `dashboard.html` and `(t) =>` in `public/js`. The `r.json()` negative assertion beside it got the same treatment for a sharper reason — pinned to one spelling it would quietly stop catching the shortcut it exists to forbid.

## Where #415 stands

| target | limit | now |
|---|---|---|
| inline JS | 2,000 | **6,281** |
| inline CSS | 800 | 4 (met) |
| `public/css/dashboard.css` | 800 | split into 8 files, largest 489 (met) |

**One extraction was tried and reverted, on purpose.** The "Cross-case capture warning" block — 1,025 lines, one cohesive cluster, three cleanly-fixable escapes — passes every filter the inventory has and is the page's case-load path: `connect()`, the state save every refresh depends on, and the refresh fan-out. Two lifecycle gates rejected it by name. Cohesion measures whether functions call each other, which core machinery does more tightly than any feature, so the inventory cannot have an opinion here. That is now written beside the cohesion pass, because the next person meets this block at the top of the queue with the same clean-looking numbers.

**The mechanical phase is over.** 34 sections and 8,084 lines remain, and only one 7-line section passes every filter. What blocks the rest is state: a cache, a snapshot, a marker shared across a boundary. Two worked examples now exist. Case templates closed one escape with an accessor. Velociraptor triage closed **seven**, and needed three different answers: three bindings were simply declared in the wrong file, one had its ownership backwards (thirteen uses and two assignments in a module, one reset in the page), and three are genuinely shared and stay behind accessors.

**A third answer, and the cheapest.** `CP_MARK` and `CP_LABELS` were frozen collection-plan tables declared under the *keyboard-navigation* banner, each with exactly one reader, and both readers already modules. Not escapes — two `const`s in the wrong file. Moving them took that section from two escapes to zero and made 134 lines extractable.

**Elements do not determine ownership; the state they touch does.** Three times now a binding's owner was not the feature its element belongs to: the enrichment overlay inside the search block, the velo monitor controls inside the bundle builder, and `#miValue`'s Escape handler, which reads as manual-IOC-form wiring until you notice it clears the one-shot recording whether the *keyboard* opened that form.

**An accessor boundary has a failure mode a shared binding does not.** The facade stubs an accessor to a no-op returning `undefined`, and readers do `.find()` on the result — a TypeError, the exact thing the facade exists to prevent. Stubbing a getter is not the same as stubbing a command, and every remaining state fix will hit this.

The case-templates extraction is the first worked example of the fix — `_cachedTemplates` was written from another module across the shared global lexical environment, and the owner now exposes `invalidateTemplateCache()` instead. That is #415's item 2, one feature at a time, and it is what the remainder needs.

**The easy extractions are done.** Every section still in the queue is flagged as holding more than one feature, so the remaining work is splitting banners rather than moving them — the finding-workflow / pinned-findings pair is the first of that shape and shows why it matters: shipped as one module, a missing file would have taken both features down and named the wrong one in the chip.

Items 1 and 3 are done. Item 2 is open and is the long tail — this PR does not finish it, and no single PR can. What it does is replace "pick a block and hope" with a measured queue.

**It measures cohesion.** Three banners in a row held two features under a heading naming only one — worst was "Push ingest token (#84)", 222 lines of which 47 are the push token and the rest are the Velociraptor bundle builder belonging to the feature *above* the banner. Connected components over the block's own declarations now catch that. Validated against the pre-extraction file, where it reports `[10, 5]` — exactly the split I had found by hand. 32 of 80 sections flagged. It is a reading prompt, not a verdict — and it has already been wrong once: `doAsk`, the AI Ask box, sat under the "Import undo / redo" banner as a component of one, and the first version dropped singletons as noise. A lone lookup table is noise; a lone forty-four-line function is a feature. Fixed and pinned, and re-validated against the file as it stood before that extraction.

**It finds the stanzas earlier extractions left behind.** When a feature moves out, its `if (typeof initX !== "undefined") …` guard stays between two banners, so the inventory files it under whichever section encloses it. NSRL's range ended with six lines belonging to the Settings → Tools extraction, and copying the range wholesale put a call to the page's `dfirFeatureUnavailable` inside a module. There are **fifteen sections** carrying one — every extraction in this PR has left one, and each is a trap for whoever takes that section next. The first version of the detector required the literal `undefined` and missed `=== "function"`; that gap let a stanza travel into the asset-overrides module and stop the ticket integrations initialising, silently. Widening it took the count from 7 sections to 15.

**And the census stopped demanding renames.** "leaves nothing behind" compared bindings at any depth on **both sides**, which cost three renames the code did not want — `mcpTick`, `wireToolRules`, `pollHuntResults` — and was about to demand a fourth for `const v = (elId, val) => …`, because three unrelated page functions each name a local `v`. The page side is top-level only now: that is what a later script can see and what "left behind" can mean. Loosening a gate to pass one's own change deserves proof, so both directions are pinned — a top-level leftover still fails, the nested namesake no longer does.

Open follow-ups this series produced: #476, #477, #479, #480, #482, #489.

Refs #415


hasamba:
Next step, found while queueing the following extraction — recording it so it is not re-derived.

**The Kill Chain feature is split across two banners.** Its constants and detail pane (`KC_TACTIC_PRIORITY`, `KC_CHAIN_ORDER`, `KC_SEV_ORDER`, `kcSelectedTac`, `kcByTactic`, `kcSelect`, `kcRenderDetail`) sit at roughly `dashboard.html:5845-5908`, under the **"Screenshot OCR full-text search"** banner. `renderKillChain` is at **6598**, under a different one, and `data-act="kcSelect"` dispatches from 6622.

That split is the entire reason the 194-line OCR section reports five state escapes. All five are Kill Chain names read by Kill Chain code. Reunite the two halves and the OCR section is ready with zero escapes — the same move that unblocked keyboard-nav, where `CP_MARK` / `CP_LABELS` went to their sole readers.

Two things to check before moving anything:

1. **Where the stanza actually starts.** `KC_TACTIC_PRIORITY` is at 5845, but the regex array closing at 5844 is read from inside the Kill Chain matcher at 5872 and is probably Kill Chain too. Find its opening line before choosing a range, and run `npm run split:dashboard-section` on the result — it refuses a range that cuts a statement in half.
2. **`KC_SEV_COLOR` is genuinely shared** and should NOT travel with the rest. Three extracted modules read it by bare name — `dashboard-anomalies.js`, `dashboard-timeline-gaps.js`, `dashboard-derived-panels.js` — plus the page at 6177 and 7321. It is a severity palette, not Kill Chain state. It needs its own answer (a shared owner that publishes it), separate from this move.

hasamba:
Next after the Kill Chain move — traced, not started, so it is not re-derived.

**"Second LLM opinion (#116)"**, `dashboard.html:5315-5425` (111 lines, 5 escapes). Three of the five are read from exactly one place, and it is the wrong place: `public/js/dashboard-search-scope.js:305-318` holds the delegated click handler for `#secondOpinionPanel`, which reads `soCollapsed`, `SO_COLLAPSE_KEY` and `lastSecondOpinionRec`. That handler is the second-opinion feature's own wiring; it was swept into search-scope during an earlier extraction.

That makes five for the tally this PR has been keeping — enrichment overlay in the search block, velo monitor controls in the bundle builder, `#miValue` Escape in keyboard nav, `sessionsCollapsed`, and now this. **An element's address is not its owner; the state it touches is.** The difference here is that the wrong address is another *module*, not the page, so the census gates could not see it — a bare-name read from a sibling module resolves through the shared global lexical environment and looks local to nobody.

**The shape of the work:**

1. Extract `5315-5425` as `dashboard-second-opinion.js`, taking the `#secondOpinionPanel` listener out of `dashboard-search-scope.js` and making it the initializer. Same two-source pattern as `dashboard-evidence-graph.js` — and run `npm run split:dashboard-section` on the body range, which I skipped there and it cost a silently-empty graph.
2. The other two escapes need accessors, not moves. `fpAiConfigured` is read at `dashboard.html:4737` and `8476`, `secondOpinionEnabled` at `8473` and `8475`. Both are read by page code that stays, so both cross as questions.
3. Check the facade direction on those two. If a stub answers "not configured" / "disabled", the page hides the feature — a decoration lost, which is the safe way round. Confirm before stubbing.

hasamba:
**"Attack Phases"** (`dashboard.html:5629-5956`, 328 lines) is two features under one banner, and should be split before either moves.

Only `phasesData` is Attack Phases. The other four escapes are **IOC provenance / corroboration state**: `iocSourcesById` (5650), `iocProvenanceFilter` (5672), `riskIocsFilter` (5709), `iocProvenanceChains` (5739). They are filed under the Attack Phases banner and have nothing to do with it.

**Why they cannot simply move with a module.** Their reader is `renderIocs` — `dashboard.html:6783-6785`, three consecutive filter lines in the page's core IOC render path:

```
if (DfirTimelineView.corrobIocs() > 1) visible = visible.filter(i => (iocSourcesById[i.id] ...
if (iocProvenanceFilter !== "all")     visible = visible.filter(i => iocProvenanceOf(i.id) ...
if (riskIocsFilter > 0)                visible = visible.filter(i => iocRiskRankOf(i.id) >= ...
```

`renderIocs` is named core machinery and stays in the page. So all four cross as **questions, not variables** — the same answer as `isNewEvent`/`isNewIoc` in extraction 79. Watch the facade direction carefully here, because it is the opposite of the usual one: these are *filters*, so a stub answering falsy means **no filtering**, i.e. the analyst sees MORE rows than they asked for. Showing un-filtered evidence is safer than hiding evidence, but it must be deliberate and the panel should say so — a silently-inactive filter is a lie about what is on screen.

`riskIocsFilter` is additionally read by `dashboard-search-scope.js`, which is the sibling-module read that no census gate can see (see extraction 82).

**Suggested order:** split the banner first as its own commit — Attack Phases keeps `phasesData`, the four IOC bindings become their own section — then extract each half. Splitting a banner is cheap and reversible; untangling two features mid-extraction is neither.

hasamba:
**Last section with real feature code in it: "Starred events" (`dashboard.html:4805-4929`).** Traced, not started.

Eight of its 125 lines are other features' guard stanzas (`4895-4928`), so the feature is ~90 lines. The split script refuses the full range — use `4805-4894`.

**Five escapes, and they are all the same kind — IOC noise-reduction toggles read by `renderIocs`:**

| binding | read by |
|---|---|
| `hideFpNoIntel` | `dashboard-search-scope.js` |
| `hideSystemPaths` | `dashboard-search-scope.js` |
| `showSignalIocsOnly` | `dashboard-search-scope.js` |
| `showFlaggedIocsOnly` | page only |
| `starredTagIds` | `dashboard-bulk-select.js` |

Three of the five are read by `search-scope` because **the toggle controls live there** — the filter bar. That is the eleventh time in this PR that a control's handler was in one file and its state in another; it is the same finding as `riskIocsFilter` (extraction 84) and the `#secondOpinionPanel` handler (extraction 82).

**Do this the way IOC provenance was done, not the way it looks.** These are filters, so the facade direction inverts: a falsy stub means NO filtering and the analyst sees more rows, not fewer. Safer than hiding evidence, but it has to be visible. Prefer ONE exported operation that applies all the active lenses and returns the list, plus one `...FiltersActive()` for the empty-state text, over five separate predicates — that is what made extraction 84 come out clean.

**And check `starredTagIds` separately.** It is the only one of the five that is not a filter toggle, and `dashboard-bulk-select.js` reads it. Work out whether the star set belongs to this feature or to selection before giving it an accessor.

hasamba:
**Correction to my own queue reading, and the work it was hiding.**

I filtered the extraction queue on `sharedMachinery` as though it were a blocker. It is not — it flags a function several sections call, which is an argument about what to PUBLISH, not about whether a block can move. That filter hid **seven sections, ~1,300 lines**, all still extractable:

| lines | section | escapes | note |
|---|---|---|---|
| 337 | Login Graph (cytoscape) | 2 | **split the banner first** — see below |
| 285 | Comprehensive Setup wizard (#181) | 7 | 11 foreign guard stanzas in range |
| 218 | Background jobs (#225) | 1 | |
| 119 | Analyst triage tags | 2 | |
| 101 | Compromised assets + asset↔IoC graph | 4 | entangled with Login Graph |
| 94 | Theme (#53) | 3 | |
| 15 | Suggested Fleet Hunts (#57) | 4 | |

**Login Graph needs a banner split before extraction, and this is the trap.** Its banner (`4227-4563`) contains TWO graph features: the login graph (`lgGV`) and the asset graph (`assetGV`, `assetEnsureGV`, `renderAssetGraph`, `renderAssetList` — roughly `4505-4563`). The asset graph is used by the separate "Compromised assets" section that follows it. Moving the banner wholesale takes one feature's state away from another.

Their wiring is in a third place — the page's load-time block at `7549-7561`, three statements:

- `7549-7552` `.asset-type-toggle` change handlers — the ASSET graph's
- `7554-7556` `#sec-assets h2` expand → `assetGV.onExpand()` — the ASSET graph's
- `7559-7561` `#sec-login-graph h2` expand → `lgGV.onExpand()` — the LOGIN graph's

Stop before `7564`, which is the evidence graph's guard stanza.

**Suggested order:** split the banner between the two graphs first (cheap, reversible, no code moves), confirm the inventory then reports two sections with their own escapes, and only then extract each with its own slice of `7549-7561`. This is the same shape as the Kill Chain and Attack Phases splits earlier in this PR — both of which were only safe because the banner was fixed before anything moved.

hasamba:
**"Comprehensive Setup wizard" (`dashboard.html:7699-8056`) is a fourth stale-banner repair, not an extraction.**

Both wizard modules already exist — `js/dashboard-setup-wizard.js` and `js/dashboard-wizard-ai-step.js`. `initSetupWizard` is itself one of the **21 foreign guard stanzas** in this range. What stayed behind is the state, exactly as it did for pinned findings, super-timeline and anonymization.

**All seven escapes are read by those two modules and nothing else:**

| binding | line | reader |
|---|---|---|
| `WIZ_DISMISS_KEY` | 7705 | setup-wizard |
| `WIZ_MODEL_HINTS` | 7706 | **wizard-ai-step** |
| `LOCAL_PROVIDERS` | 7714 | **wizard-ai-step** |
| `WIZARD_BY_ID` | 7810 | setup-wizard |
| `WIZ_ORDER` | 7809 | setup-wizard |
| `wizCurrent` | 7811 | setup-wizard |
| `wizStatus` | 7812 | setup-wizard |

**Two things make this harder than the earlier three repairs:**

1. **They interleave.** The two wizard-ai-step bindings (7706, 7714) sit *between* the setup-wizard ones (7705, and 7809-7812), and `WIZARD_STEPS` (7722) — which `WIZ_ORDER` and `WIZARD_BY_ID` are both derived from — is built with `F`, the field-descriptor factory at 7721. `F` has 31 call sites, all inside this range (confirmed via the new `vocabulary` field), so it travels with `WIZARD_STEPS`.

2. **`WIZ_MODEL_HINTS` and `LOCAL_PROVIDERS` cannot simply be published.** They are a lookup object and a Set, and the feature manifest requires every published name to be callable — that gate rejected a bare `QA_AUDIT_MARK` in extraction 76 for the same reason. They need accessors: `wizModelHints(provider)` and `isLocalProvider(name)`, owned by whichever module ends up holding the data.

**Suggested order:** move the setup-wizard group (`WIZ_DISMISS_KEY`, `F`, `WIZARD_STEPS`, `WIZ_ORDER`, `WIZARD_BY_ID`, `wizCurrent`, `wizStatus`) into `dashboard-setup-wizard.js` first and confirm the inventory drops to two escapes; then move `WIZ_MODEL_HINTS` and `LOCAL_PROVIDERS` into `dashboard-wizard-ai-step.js` with accessors if anything outside still needs them. Do not try both in one pass — the interleaving is what makes a single splice go wrong.

hasamba:
The scope question is now raised on the issue itself: https://github.com/hasamba/DFIR-Companion/issues/415 — it asks whether the 2,000-line target should be re-baselined to ~2,800 (the floor with the spine inline) or kept, with spine modularisation split out as its own design decision. This PR is complete and CI-green either way; only the "is it done?" answer depends on that call.

PR #490, opened 2026-08-07
docs(architecture): state what the inline-JS target measures (#415)

### LINKED ISSUE #415: refactor(dashboard): decompose dashboard.html — extract what is standalone, then restructure the shared state

hasamba:
`ARCHITECTURE.md:305` still showed **16,634** — the number from before #488 — for a target that is now at **2,155**. Updating it turned up something worth writing down instead of quietly resolving.

## Two numbers were in circulation

| measure | value | reads as |
|---|---|---|
| `check:size` total — all 5 src-less `<script>` blocks | **2,155** | not met |
| main block alone (what the inventory tool reports) | **1,967** | *met* |

Only the first has ever been the target's measure. The gap is 188 lines across four small blocks, and picking the flattering number would have closed #415 on 170 lines of extractable code.

## What the target now says it measures

All inline JS **minus the pre-paint theme bootstrap** (`:357`, 18 lines) → **2,137 today, 137 over.**

That block is the only one that genuinely cannot become an external script: it reads the stored theme and sets the attribute before first paint, so any `src=` makes it a round-trip that paints the wrong theme first. The exclusion is about what is *possible*, not a discount.

## What is deliberately not excluded

| block | lines | why it stays counted |
|---|---|---|
| team-mode `fetch` wrapper `:15` | 82 | only needs to run first — an external `<head>` script does that |
| Claude Code status card `:1343` | 45 | ordinary feature code |
| Codex status card `:1450` | 43 | ordinary feature code |

Extracting these three takes the net measure to **1,967** and the ratchet total to **1,985** — both under 2,000. Recorded in the doc as the remaining work for #415.

**The ratchet is unchanged** and still counts all five blocks, so nothing can regress while this sits.

Docs only — no code, no gate changes.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -302,12 +302,33 @@ the issue that owns the work, so no number is a blocker without an assignee:
 |---|---|---|---|
 | `analysis/pipeline.ts` | 800 | **591 — met** | [#418](https://github.com/hasamba/DFIR-Companion/issues/418) |
 | `server.ts` | 800 | **623 — met** | [#416](https://github.com/hasamba/DFIR-Companion/issues/416) |
-| `public/dashboard.html` inline JS | 2,000 | 16,634 | [#415](https://github.com/hasamba/DFIR-Companion/issues/415) |
+| `public/dashboard.html` inline JS | 2,000 | 2,155 (2,137 net — see below) | [#415](https://github.com/hasamba/DFIR-Companion/issues/415) |
 | `public/dashboard.html` inline CSS | 800 | **4 — met** | [#415](https://github.com/hasamba/DFIR-Companion/issues/415) |
 | `public/css/dashboard-*.css` (8 parts) | 800 | **489 max — met** | [#415](https://github.com/hasamba/DFIR-Companion/issues/415) |
 | Files in `src/` over 800 lines | 0 | 11 | the ledger below |
 | Flat files in `src/analysis/` | 0 | 296 | whichever extraction touches them |
 | Boundary ledger | 10 | 47 | shrinks as the above land |
+
+**What "inline JS" counts, and the one thing it does not.** `check:size` sums every `<script>` in
+`dashboard.html` that has no `src` — currently five blocks totalling **2,155** lines. That total is
+the ratchet, and it stays the ratchet: it can only shrink.
+
+The 2,000 target is measured against the same total **minus the pre-paint theme bootstrap**
+(`dashboard.html:357`, 18 lines) — **2,137** today. That block is the only one on the page that
+cannot become an external script: it reads the stored theme and sets the attribute before first
+paint, and any `src=` turns it into a network round-trip that shows the wrong theme first. Excluding
+it is a statement about what is *possible*, not a discount on what is left to do.
+
+Nothing else is excluded, deliberately. The other three non-main blocks — the team-mode `fetch`
+wrapper (`:15`, 82 lines) and the two AI-provider status cards (`:1343` and `:1450`, 88 lines) — are
+ordinary code that happens to sit inline. The wrapper only needs to run first, which an external
+script in `<head>` does equally well. Counting the main block alone would have read **1,967 — met**,
+and that number is why this paragraph exists: it reaches the target by exempting 170 lines of
+extractable feature code, which is a different thing from meeting it.
+
+**Still 137 lines over.** Extracting the wrapper and the two status cards leaves the main block plus
+the bootstrap: the net measure becomes **1,967** and the ratchet total **1,985**, so both land under
+2,000 and #415 closes honestly. That is the remaining work, not a rounding argument.
 
 **The ratchets hold every one of these flat in the meantime.** `check:size` freezes each file at its
 recorded length, `check:imports` at one known cycle, `check:boundaries` at the recorded violations.


hasamba:
**Superseded by #504.** Closing — but the finding in here was right, and #504 carries it forward.

## What held up

The core observation stands: two numbers were in circulation for one target, and `ARCHITECTURE.md` named neither. That is now stated explicitly — all five src-less `<script>` blocks, which is what `check:size` ratchets, currently 1,776 + 82 + 45 + 43 + 18 = **1,964**.

The warning that *"picking the flattering number would have closed #415 on 170 lines of extractable code"* was worth making. It did not happen: #501 measured on the strict ratchet total throughout, and the ledger records the same number the gate does.

## What the work overtook

Both of the mechanisms proposed here are no longer needed, because the target is met without them:

- **Excluding the 18-line pre-paint theme bootstrap.** That exclusion existed to bring an unreachable target within reach. 1,964 is under 2,000 with the block counted, so introducing an exclusion now would only weaken a target already met.
- **Extracting the team-mode fetch wrapper and the two status cards** (82 + 45 + 43) to reach 1,985. #501 took 193 lines out of the main block instead — comments, report meta, confidence control and the activity log, each into its own module — landing at 1,964.

So the three small blocks stay inline, and the ratchet still counts them. Nothing regresses while they do.

## Note for anyone filing a docs-only PR here

This sat open since 7 August because branch protection requires status checks that the path-filtered workflows never trigger on a docs-only change — `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED`, no checks reported. #504 hit the same wall. Worth either adding a docs path to the workflow triggers or a required-check exemption, or every documentation fix needs an admin merge.

PR #493, opened 2026-08-08
fix(test): walkLoadTime follows calls, not just IIFEs (#476)

### LINKED ISSUE #476: test(dashboard): the missing-module gate misses helpers invoked at load, so it passes a page that dies
> **Updated 2026-08-05.** The defect below is unchanged and still open. What changed is the *size of the remediation* — see "What this now costs". Two gates have landed since this was filed that make most of the original tail unnecessary.

`walkLoadTime()` in `companion/tests/helpers/dashboardAst.ts` follows IIFEs that are invoked directly, but not an ordinary helper function that is *called* during page load. Everything reachable through that one hop is invisible to the missing-module gate.

## This is not theoretical — it shipped

PR #475 added the derived `ENTRY_POINTS` gate and its whole suite went green. Blocking `/js/dashboard-timeline-view.js` in Chromium anyway produced:

```
ReferenceError: DfirTimelineView is not defined
    at renderExcludeChips (dashboard:3622)
    at dashboard:15814
```

`renderExcludeChips()` holds a bare reference and is called from a top-level statement. One hop, invisible to the gate, and the throw aborts the rest of the inline script exactly as a direct reference would.

## Reproducing the hole

Replace the real guarded call with this and every relevant test still passes:

```js
function boot() {
  initTicketIntegrations();
}
boot();
```

Also missed: `window.N()`, `window["N"]()`, `.call()`-form IIFEs, and synchronous callbacks.

## Scale of what it cannot see

Running the repo's own `buildCallGraph()` + `reachableFrom()` over the inline script found **38** load-reachable bare references to `DfirTimelineView` alone, plus 17 module-published names called bare on load-time paths.

The total is method-dependent and four different measurements disagreed on the digit (365/42, 295/80, 276/64, 270/62 refs/names) while agreeing on the shape — a few hundred. One of them was inflated by 969 references to `esc`/`escAttr`, which the inline script declares itself and which can never throw. **Whoever picks this up should publish one number with its method before estimating anything.**

## What this now costs — the part that changed

This issue originally implied a large tail: fix the helper, then guard, stub or exempt each of the few hundred references it starts reporting. That tail is mostly gone, because two merge-time gates now close the road it was defending:

- **`settingsSearch.test.ts:161`** — every `<script src="/js/…">` the dashboard loads must be in `STATIC_ASSETS`. A module that is loaded but never registered cannot reach master.
- **`dashboardNameResolution.test.ts`** (#487) — every name the page reads must be declared in the page, published by a module it loads, or a browser built-in. A typo, a rename and a deletion cannot reach master either.

Together those mean a missing file at runtime is a **corrupt install**, not a build mistake — the server binds `127.0.0.1` and reads these off local disk (`src/server.ts:487`). So the remediation is no longer "a decision per reference". It is one decision:

> **What should the page do when its own install is incomplete?**

"Name the missing file and stop" is a defensible answer for all of them, and it beats both several hundred `typeof` guards and several hundred stubs. Note `check:size` has zero headroom on `dashboard.html#inline-js`, so any answer that *adds* inline lines needs either line-neutrality or an `--init` re-baseline with justification in the PR.

There is still an unresolved policy underneath this, and it should be settled as part of the work rather than discovered during it: `public/js/dashboard-facade.js` argues stubs beat guards because guards rot, while this gate's failure message prescribes `typeof` guards, and neither knows the other exists. They have never collided only because the gate is blind.

## Done looks like

1. `walkLoadTime()` follows the call graph transitively, the way `calleesInsideLoops()` already does for the loop checker — that gate's own test is literally *"no function reachable from a loop through any number of hops commits"*, so the machinery exists in this file already.
2. Contract tests in `dashboardAstContract.test.ts` for each shape above, each mutated to prove the gate fires. This repo's standard is mutation evidence, not a green suite.
3. One documented answer to the corrupt-install question, applied uniformly, with the facade-vs-guard policy written down.

## Already landed, do not redo

- **#478** — `unguardedTopLevelRefs()` discarded shorthand properties (`{ initTicketIntegrations }` evaluates the binding). Fixed in #484; it shares this function's call path, which is why it went first.
- **#487** — the name-resolution gate described above.
- The tier-3 page-breakers themselves (#475, `c1f3e3ed`): the temporal-dead-zone abort in `dfirFeatureUnavailable`, the unguarded refresh fan-out, and `DfirTimelineView`'s deliberate halt.

Found by an independent audit of PR #475 at `cfdef75e`, and separately by the blocked-script Chromium pass on the same commit.

hasamba:
Closes #476.

`walkLoadTime()` stopped at the syntactic load-time region — top level plus IIFE bodies — so an ordinary helper **called** during load was invisible. One hop was enough to ship a page that dies: `renderExcludeChips()` held a bare `DfirTimelineView` reference and a top-level statement called it, so blocking that module aborted the whole inline script while #475's suite stayed green.

## The fix

`walkLoadTime` now takes a fixpoint over what load-time code calls — the same shape `calleesInsideLoops()` already uses for the loop checker, whose own contract test is *"no function reachable from a loop through any number of hops commits"*.

Also widened to the spellings the page uses and the gate did not recognise:

| Shape | Why it was missed |
|---|---|
| `boot()` where `boot` is declared elsewhere | the one-hop hole itself |
| `window.N()` · `window["N"]()` | callee is not a bare identifier |
| `f.call(…)` · `f.apply(…)` | the invoked function is the *object*, not the property |
| `(function(){…}).call(this)` | `.call`-form IIFE — the old `iifeBodyOf` only matched the direct-invocation shape |
| `ready(handler)` | a by-name hand-off; there is no call expression naming `handler` anywhere |

A function merely **defined** at load is deliberately still not reported — it runs later, on an event, where a throw is contained to one interaction.

## The number, with its method

#476 asked for one measurement before anyone estimates the remediation, because four earlier attempts disagreed on the digit (365/42, 295/80, 276/64, 270/62).

**Method:** entry set = declared functions invoked from the syntactic load-time region; closure = `reachableFrom(buildCallGraph(...))`; counted = reads of a module-published name inside those bodies, excluding names the inline script declares itself — that exclusion is what inflated one earlier count by 969 `esc`/`escAttr` references.

> **Result: ZERO.** `tests/dashboard` + `tests/architecture` are **2025 green** with the hole closed.

The remediation tail this issue anticipated is gone *entirely*, not "mostly". #488's decomposition moved the reachable code out of the inline script, which is now 1,967 lines rather than 7,666 — the issue's own repro cites `dashboard:3622` and `dashboard:15814`, line numbers that no longer exist.

## What is deliberately left open

The third "done" item — *"what should the page do when its own install is incomplete?"* — is **recorded, not guessed**. With no instances to apply it to, it is a decision about product behaviour with nothing currently riding on it, and the conflict it names (`dashboard-facade.js` argues stubs beat guards; this gate's message prescribes `typeof` guards) has still never collided. Deciding it here would be inventing a policy for a hazard that does not presently exist.

## Mutation evidence

Per this repo's standard, the new tests were proven to have teeth rather than assumed. Disabling the fixpoint fails **6 of the 9** new contract tests. The 3 survivors are the ones that should survive: two controls that must stay silent (defined-but-not-called, `typeof`-guarded) and the `.call()`-form IIFE, which the separate `iifeBodyOf` extension covers.

Each positive case is paired with its silence control, so a future regression in either direction shows up: if the walk goes back to stopping at the region the first half goes quiet; if it starts entering every function body the second half starts shouting.

## Checks

`typecheck`, `lint`, `format:check` clean · `tests/dashboard` + `tests/architecture` 2025 passed

### DIFF
--- a/companion/tests/dashboard/dashboardAstContract.test.ts
+++ b/companion/tests/dashboard/dashboardAstContract.test.ts
@@ -741,3 +741,25 @@ describe("#476 — a module reference one call from load is a load-time hazard",
     ).toHaveLength(1);
   });
 });
+
+// Codex review of the sibling #477 change found the same unrestricted branch here: a property call
+// resolved to ANY same-named declared function. Following `api.boot()` into a local `boot` walks a
+// body that never runs at load, so the gate reports hazards that cannot happen — the false-alarm
+// direction, which is how a gate loses its readers.
+describe("#476 — a property call only reaches a page function through a global root", () => {
+  const MISSING = new Set(["DfirTimelineView"]);
+  const linesFor = (src: string): number[] =>
+    unguardedTopLevelRefs(scriptFromSource("p.js", src), MISSING).map((r) => r.line);
+
+  it.each([
+    ["an unrelated method of the same name", `function boot() { DfirTimelineView.x; }\napi.boot();`],
+    ["a computed member of a non-global", `function boot() { DfirTimelineView.x; }\napi["boot"]();`],
+  ])("does not follow %s", (_label, src) => {
+    expect(linesFor(src)).toHaveLength(0);
+  });
+
+  it("still follows the global-root spellings", () => {
+    expect(linesFor(`function boot() { DfirTimelineView.x; }\nwindow.boot();`)).toHaveLength(1);
+    expect(linesFor(`function boot() { DfirTimelineView.x; }\nglobalThis["boot"]();`)).toHaveLength(1);
+  });
+});
--- a/companion/tests/dashboard/dashboardAstContract.test.ts
+++ b/companion/tests/dashboard/dashboardAstContract.test.ts
@@ -660,3 +660,84 @@ describe("duplicateBindings compares the page at its top level", () => {
     expect(duplicateBindings("dashboard-thing.js", moduleSrc, page(pageSrc))).toEqual([]);
   });
 });
+
+// ── #476: THE LOAD-TIME WALK FOLLOWS CALLS, NOT JUST IIFEs ──────────────────────────────────────
+//
+// walkLoadTime() used to stop at the syntactic load-time region — top level plus IIFE bodies — so an
+// ordinary helper CALLED during load was invisible. One hop was enough to ship a page that dies:
+// renderExcludeChips() held a bare DfirTimelineView reference and a top-level statement called it,
+// so blocking that module aborted the whole inline script while every gate stayed green.
+//
+// Each case below is written the way the page writes it, and each is paired with the DEFINED-BUT-
+// NOT-CALLED twin that must stay silent. That pairing is the mutation evidence: if the walk ever
+// goes back to stopping at the region, the first half goes quiet; if it starts entering every
+// function body, the second half starts shouting.
+describe("#476 — a module reference one call from load is a load-time hazard", () => {
+  const MISSING = new Set(["DfirTimelineView"]);
+  const linesFor = (src: string): number[] =>
+    unguardedTopLevelRefs(scriptFromSource("p.js", src), MISSING).map((r) => r.line);
+
+  it("follows an ordinary helper invoked by a top-level statement", () => {
+    // The exact reproduction named in #476.
+    expect(linesFor(`function boot() { DfirTimelineView.hydrate({}); }\nboot();`)).toHaveLength(1);
+  });
+
+  it("stays silent when that same helper is only DEFINED", () => {
+    // The control. A function that merely exists runs later, on an event, where a throw is
+    // contained to one interaction — reporting it would bury the real hazards in noise.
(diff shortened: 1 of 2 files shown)


hasamba:
Carried a finding over from the Codex review of the sibling #477 change (#494): `invokedNames` had the same unrestricted property-call branch, so `api.boot()` resolved to the page's local `boot` and the walk entered a body that never runs at load — reporting hazards that cannot happen.

The harm runs the opposite way to #477's (false alarms rather than missed hazards), but it is the same mistake, and a gate that cries wolf is how a gate loses its readers. A property call now resolves through a global root only (`window`/`globalThis`/`self`), with four tests pinning both directions.

`tests/dashboard` + `tests/architecture` **2028 passed**; typecheck, lint, format:check clean.

PR #494, opened 2026-08-08
fix(test): callsByName needs an invocation, not a mention (#477)

### LINKED ISSUE #477: test(dashboard): callsByName() treats any top-level mention as reachability, so an uncalled initializer passes

hasamba:
Closes #477.

`callsByName()` drew its reachability edges from any identifier **read**, which answers *"is this name mentioned here"* — a different question from *"is this initializer reached"*.

```js
function dead() {
  initTicketIntegrations();
}
void dead;          // ← a mention. `dead` was marked live; the call inside it counted.
```

A feature could be fully unwired — no listeners, no fetches, no published entry point — with the suite green. This is the check standing between the repo and an initializer that silently stops running.

## What an edge needs now

| Accepted | Why |
|---|---|
| `f()` · `(f)()` · `f.call(…)` · `f.apply(…)` | a call, in the spellings the page uses |
| `window.f()` · `window["f"]()` | same call, through the global object |
| `el.addEventListener("click", h)` · `setTimeout(h, 0)` · `xs.forEach(h)` · `ready(h)` | a hand-off **by name** — how every registration spells it |
| `el.onclick = h` | event-handler property assignment |

| Rejected | |
|---|---|
| `void dead` · `dead;` · `const keep = dead` · `[dead]` · `dead === undefined` · `typeof dead` | a bare read invokes nothing |

## It also fixed two false negatives nobody had filed

`window.live()` and `window["live"]()` never created an edge under the old rule either — the callee is not a bare identifier, so a page whose entry point is invoked through the global object read as **unreachable**. Both are now covered, and both are pinned by tests.

## The honest limit

A name passed to **any** call counts as invocable, because telling `ready(h)` from `console.log(h)` requires knowing what the callee does with it. That is written into the helper rather than left for someone to rediscover. Erring that way keeps false alarms out of a gate whose entire value is that people trust it when it fires.

## Mutation evidence

Restoring the mention-based edge fails **7 of the 16** new contract tests — five that must stay silent, plus the two window-spelling cases that must fire. Both directions are pinned, so a future regression either way shows up.

## Relationship to #476

#477 suggested reusing #476's reachability if that landed first. It has not (#493 is open), so this is independent and based on `master`. The two touch the same file but different functions — `callsByName` here, `walkLoadTime` there — and whichever merges second will need a trivial rebase. The call-spelling logic is deliberately named `invocationEdges` here to avoid colliding with #493's `invokedNames`; folding them into one helper is a good follow-up once both are in.

## Checks

`typecheck`, `lint`, `format:check` clean · `tests/dashboard` + `tests/architecture` **2032 passed**

### DIFF
--- a/companion/tests/dashboard/dashboardAstContract.test.ts
+++ b/companion/tests/dashboard/dashboardAstContract.test.ts
@@ -692,7 +692,24 @@ describe("#477 — a mention is not a call", () => {
     ["setTimeout by name", `function h() { N(); }\nsetTimeout(h, 0);`],
     ["forEach by name", `function h() { N(); }\n[1].forEach(h);`],
     ["handed to a helper", `function h() { N(); }\nfunction ready(f) { f(); }\nready(h);`],
+    // Codex review of the first cut: an alias IS an invocation path. Dropping the `run -> live`
+    // relationship reported a genuinely-reached initializer as unreachable — a false alarm, which
+    // is the failure mode that teaches people to ignore the gate.
+    ["an invoked alias", `function live() { N(); }\nconst run = live;\nrun();`],
+    ["an invoked alias assigned later", `function live() { N(); }\nlet run;\nrun = live;\nrun();`],
   ])("still counts %s", (_label, src) => {
     expect(cb(src)).toBe(true);
+  });
+
+  // Also from that review, and the more dangerous direction: a property call only resolves to a
+  // page-level function through a GLOBAL ROOT. `api.dead()` is somebody else's method that happens
+  // to share the spelling, and counting it revived a function nothing invokes — reintroducing this
+  // very issue by a different door.
+  it.each([
+    ["an unrelated method of the same name", `function dead() { N(); }\napi.dead();`],
+    ["a computed member of a non-global", `function dead() { N(); }\napi["dead"]();`],
+    ["an alias that is never invoked", `function live() { N(); }\nconst run = live;`],
+  ])("does not count %s", (_label, src) => {
+    expect(cb(src)).toBe(false);
   });
 });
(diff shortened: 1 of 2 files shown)


hasamba:
Ran `/codex:review --scope branch --base origin/master` against this branch. It returned two P2 findings; **both were real, both reproduced, both are now fixed** in `d0e9e1a`.

**1. A property call resolved to any same-named local.**

```js
function dead() { N(); }
api.dead();                 // ← marked the local `dead` live
```

This reintroduced the exact defect the PR is about, through a different door — a function nothing invokes became reachable. A property call now resolves to a page-level function **only through a global root** (`window`/`globalThis`/`self`), where `window.live()` genuinely *is* that function. Same restriction on the computed form (`api["dead"]()`).

**2. An invoked alias was dropped.**

```js
function live() { N(); }
const run = live;
run();                      // ← reported `N` as unreachable
```

A false alarm — the failure mode that teaches people to ignore a gate. Aliases are now graph *edges* rather than calls, which keeps both halves right: invoking the alias reaches the target, while `const run = live;` on its own leaves both unreachable.

Five more contract tests, including the never-invoked-alias control that stops the alias edge from quietly making everything reachable.

`tests/dashboard` + `tests/architecture` **2037 passed**; typecheck, lint, format:check clean.

Worth noting for whoever reviews #493: finding 1 applies there too. Its `invokedNames` has the same unrestricted property-name branch — the harm runs the other way (over-approximation, so false alarms rather than missed hazards), but it is the same mistake and I'll fix it on that branch.

PR #497, opened 2026-08-08
docs: state the real Node floor, and gate it against package.json (#474)

### LINKED ISSUE #474: Docs state a Node 20 floor, but the server requires Node 22.5+ to start
## Summary

`USER_MANUAL.md` tells new users to install **Node 20**, but the server throws on startup below **Node 22.5**. Anyone following the install instructions on Node 20 gets a broken install.

## The mismatch

`companion/package.json:7-9` declares the real floor:

```json
"engines": {
  "node": ">=22.5"
}
```

But `USER_MANUAL.md:66` says:

> 1. Install [Node.js](https://nodejs.org/) **20 or later** (Node 22.5+ if you want the NSRL SQLite backend).

This frames 22.5 as optional — needed only for an opt-in NSRL feature. That is no longer true.

## Why 22.5 is a hard floor, not a feature flag

`companion/src/analysis/sqliteRuntime.ts` reaches `node:sqlite` (Node 22.5+) and says so in its own comment:

> The indexed case store is now the primary backend, so this is a startup/runtime floor rather than merely an optional NSRL feature.

`loadDatabaseSync()` throws:

```
node:sqlite is unavailable — DFIR Companion requires Node 22.5 or later to open case storage
```

and it is imported by core paths, not just NSRL:

- `companion/src/server.ts:359`
- `companion/src/analysis/caseSqliteWorker.ts:2`
- `companion/src/auth/authStore.ts:5`
- `companion/src/analysis/jobLedgerWorker.ts:2`
- `companion/src/analysis/nsrlDb.ts:13` (the only genuinely optional consumer)

So case storage, auth, and the job ledger all require it.

## Failure mode

There is no `.npmrc` setting `engine-strict=true`, so `npm install` only **warns** about the engines mismatch rather than failing. The install appears to succeed and the error surfaces later, at server startup, when case storage is opened — which is a worse experience than a clean install-time failure.

## Suggested fix

1. `USER_MANUAL.md:66` — state Node 22.5+ as the minimum to run at all; drop the "20 or later" and the NSRL-only framing.
2. `README.md:432` — currently scopes 22.5+ to "Indexed SQLite case storage ... (requires Node 22.5+)". Since that store is now the primary backend, promote this to a top-level prerequisite instead of a feature note.
3. Consider adding `engine-strict=true` to an `.npmrc` so the mismatch fails at install time with an actionable message rather than at first case open.

Already correct, no change needed: `killercoda/finish.md:21` ("Or from source (Node 22.5+)").

## Related

`companion/src/node-sqlite.d.ts` notes that `@types/node` is still pinned to v20, which predates `node:sqlite` — hence the hand-written ambient declaration. Worth a bump at some point, but separate from this docs bug.

---
*Found while building a repo-wide knowledge graph; verified against the source before filing.*

hasamba:
Closes #474.

`USER_MANUAL.md` told new users to install **"Node 20 or later"**, with 22.5 framed as an opt-in needed only for the NSRL SQLite backend. That stopped being true when the indexed case store became the primary backend — `node:sqlite` is imported by case storage, auth *and* the job ledger, so the server does not start below 22.5 at all.

Anyone following the install instructions on Node 20 got a broken install.

## Changes

| File | Change |
|---|---|
| `USER_MANUAL.md` | States 22.5 as a hard floor **and why**, so the next reader doesn't re-derive it as an NSRL-only extra |
| `README.md` | Drops `(requires Node 22.5+)` from the case-storage feature bullet — the README already states the prerequisite up front, and scoping it to one feature is what made it look optional |
| `companion/.npmrc` | `engine-strict=true` |

## Why the `.npmrc` matters more than it looks

npm only **warns** on an engines mismatch. The install appeared to succeed and the error surfaced later, at first case open — a much worse first experience than a clear refusal at install time with an actionable message.

## A gate, because this drift was silent

`package.json` moved and the prose did not, and **nothing compared them**. So the floor is now derived from `engines.node` rather than hard-coded: raising the engine requirement fails this test until the prose catches up, and the failure names the file that still disagrees.

It checks the specific shape of the bug — prose telling a reader to install a *lower* version — rather than any occurrence of a number, so a doc can still say "older Node releases cannot open…" historically.

**Mutation evidence:**

| Mutation | Result |
|---|---|
| Restore the original `"20 or later"` wording | ✅ fails |
| Delete `companion/.npmrc` | ✅ fails |

Both pass again once restored.

## Deliberately untouched

- `CHANGELOG.md:315` — records what was true at that release; editing history would be wrong.
- `USER_MANUAL.md`'s two NSRL-specific *"Requires Node 22.5+"* notes — still true, just no longer the *only* reason.
- `killercoda/finish.md`, `mkdocs-docs/getting-started.md`, `companion/README.md` — already correct, and now covered by the gate.

## Checks

`typecheck`, `lint`, `format:check` clean · new gate 10 tests passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -678,7 +678,7 @@ attacker path, questions). Configure both via `.env` — see `companion/README.m
 
 ## Quick start
 
-> **Prerequisite:** [Node.js](https://nodejs.org/) **22.5 or later** (which ships with `npm`).
+> **Prerequisite:** [Node.js](https://nodejs.org/) **22.19 or later** (which ships with `npm`).
 > Check with `node --version`. Everything below uses `npm`, so no other runtime is needed.
 > Indexed case storage uses the built-in `node:sqlite` module, so older Node releases cannot open
 > cases. The portable build bundles a compatible runtime.
--- a/README.md
+++ b/README.md
@@ -429,7 +429,7 @@ All importers are **deterministic (no AI call)**, read the artifact's own timest
 - **🌍 Geographic IP map** — plot geo-located IP IOCs on an interactive Leaflet world map (severity colors, victim→attacker flows, country stats, filtering, CSV export); coordinates from the opt-in GeoIP enrichment, offline-friendly (tiles overridable)
 
 ### Ops
-- **Indexed SQLite case storage** — worker-backed, cursor-paged database replaces flat JSON case state (requires Node 22.5+)
+- **Indexed SQLite case storage** — worker-backed, cursor-paged database replaces flat JSON case state
 - **Essential / All view in Settings** — opens on a curated 43-control view instead of all ~257 fields; remembered per browser
 - **Health / Diagnostics** — **Settings → Diagnostics** one-page operator view: disk usage, case count, capture/synthesis queue, redacted AI config + live *Test AI connectivity*, importer attempts (24h/7d) + recent failures; compute-on-demand case sizes; key-free copy-to-clipboard
 - **Case Statistics panel** — per-case totals, source breakdown, and import velocity in Diagnostics
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -63,7 +63,7 @@ Choose the method that fits your setup:
 
 #### From source (recommended for development)
 
-1. Install [Node.js](https://nodejs.org/) **22.5 or later**. This is a hard floor, not an optional extra: case storage, authentication and the job ledger all use the built-in `node:sqlite` module, so the server will not start below it. Check with `node --version`.
+1. Install [Node.js](https://nodejs.org/) **22.19 or later**. This is a hard floor, not an optional extra: case storage, authentication and the job ledger all use the built-in `node:sqlite` module, so the server will not start below it. Check with `node --version`.
 2. Clone or download the repository.
 3. Run:
    ```
--- a/USER_MANUAL.md
+++ b/USER_MANUAL.md
@@ -63,7 +63,7 @@ Choose the method that fits your setup:
 
 #### From source (recommended for development)
 
-1. Install [Node.js](https://nodejs.org/) **20 or later** (Node 22.5+ if you want the NSRL SQLite backend).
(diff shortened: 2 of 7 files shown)


hasamba:
Ran `/codex:review --wait --scope branch --base origin/master`. **Two findings, both real, both verified before fixing.** The first was a footgun I introduced in this PR.

**1 — `engine-strict` would have broken supported installs.**

`undici@8.3.0` is locked at `>=22.19.0`. Turning on `engine-strict` while `package.json` declared `>=22.5` hard-fails `npm install` across **22.5–22.18** — the exact range the docs I had just written called supported. I'd have replaced a late runtime failure with an install-time one for a range I was simultaneously documenting as fine.

The declared floor was never true. This is the *same* declared-vs-actual mismatch as #474 itself, one layer down. So I raised `engines.node` to `>=22.19` and followed it through all four install docs, rather than dropping `engine-strict` — the tree genuinely requires it, and stating anything lower is precisely the bug this issue is about.

⚠️ **This changes the supported floor from 22.5 to 22.19.** Evidence-based and trivially reversible, but it's your call — say the word and I'll drop `engine-strict` instead and file the undici mismatch separately.

**2 — `parseFloat` compared versions wrong.**

`parseFloat("22.5") > parseFloat("22.19")` is `true`, so the moment the floor gained a two-digit minor the staleness check stopped flagging *"install 22.5"* as stale — and since the companion assertion only requires the floor to appear *somewhere*, a doc containing both would have passed. Component-wise compare now.

**The generalisation, which is the part worth keeping.** The gate now asserts that a machine on exactly the declared floor satisfies every locked dependency's `engines`. Two subtleties learned the hard way:

- `engines` is a **range**, not a floor. `^20.19.0 || ^22.13.0 || >=24` is satisfiable at 22.19 despite containing `>=24`; reading the first `>=` out of it flagged `eslint-visitor-keys` falsely. Uses npm's own semver rather than a regex.
- **Platform-gated optional binaries are skipped.** `@img/sharp-win32-ia32` caps at Node 20 and npm never installs it here, so it cannot fail an install.

`semver` is borrowed via `createRequire` rather than added as a devDependency — a gate about dependency floors shouldn't add one.

**Mutation evidence:** putting `engines` back to `>=22.5` fails 4 of the 11 tests, naming `undici`. typecheck, lint, format:check clean.

PR #546, opened 2026-08-10
fix(a11y): clear all 49 ledgered accessibility violations

hasamba:
First of the four grandfathered baselines. **49 → 0.**

## Why this one first

The ledger was seeded at the real count so the gate could start catching regressions immediately — the right call. But it also meant 49 real violations sat frozen, and "frozen at 49" means nobody's experience gets worse and nobody's gets better. Unlike the other three baselines (import cycle, file sizes, boundaries), these reach actual users.

## Contrast — 43 nodes, 6 colour pairs

They weren't 43 separate bugs. Fixing the light-theme **tokens** fixes them all:

| Token | Was | Now | Why |
|---|---|---|---|
| `--text-faint` | `#737d8c` | `#626a77` | 3.57:1 on `--bg-secondary`, 4.16:1 even on white — **28 of the 43 nodes** |
| `--badge-warning-text` | `#e07e00` | `#9f5900` | 2.52:1 on `--bg-secondary` — orange on a pale panel, on the "next action" cards |
| `--text-muted` | `#5a6573` | `#56616e` | 4.33:1 on `--border-color`, the background of every grey chip |
| `--tag-gray-text` | `#5a6573` | `#56616e` | same |

Each replacement is the **lightest** darkening that clears 4.5:1 against the surface it actually sits on — computed, not eyeballed — so the visual hierarchy these tokens exist to express survives. **Dark theme is untouched**: every failing pair was light-theme.

### Two that weren't token problems

- **`#scopeInfo`** hardcoded the *dark* palette's hex values unconditionally. In light theme the "scope is active" chip rendered `#ffd93b` on white at **1.37:1** — invisible, on the one indicator telling an analyst they're viewing a *filtered subset* of the evidence rather than all of it. Now uses theme tokens.
- **Secondary/cancel buttons** set only a background, inheriting white text onto light grey (**1.36:1**).

Axe flagged 2 of those buttons because only three modals are scanned open. The same markup appears on **28**, and all 28 are fixed — patching only the measured two would leave the identical bug in twenty-six dialogs.

## Select names — 6 nodes

Two selects had a visible label that was never associated → joined with `for=`, which also makes clicking the label focus the control. The other four had no label at all → `aria-label`.

## Result

```
[a11y] ok — 5 scope(s), 0 known violation(s), none new
```

The ledger is now zero on all five scopes: the gate has stopped recording debt and started enforcing a standard.

## Test plan

- [x] Axe scan re-run: **49 → 0** across all five scopes
- [x] Full E2E suite: **239 passed** — token changes touch every surface, so a targeted run wasn't enough
- [x] Ledger updated via `check-a11y.mjs --update` (reductions only; no `--init`, nothing re-baselined upward)
- [x] typecheck, eslint, prettier, file-size / import-cycle / module-boundary ratchets all pass

### DIFF
--- a/companion/scripts/a11y-ledger.json
+++ b/companion/scripts/a11y-ledger.json
@@ -1,20 +1,7 @@
 {
-  "dashboard-empty": {
-    "color-contrast": 6,
-    "select-name": 2
-  },
-  "dashboard-case": {
-    "color-contrast": 33,
-    "select-name": 1
-  },
-  "modal-enrichOverlay": {
-    "color-contrast": 1
-  },
-  "modal-anonOverlay": {
-    "select-name": 1
-  },
-  "modal-settingsOverlay": {
-    "color-contrast": 5,
-    "select-name": 2
-  }
+  "dashboard-empty": {},
+  "dashboard-case": {},
+  "modal-enrichOverlay": {},
+  "modal-anonOverlay": {},
+  "modal-settingsOverlay": {}
 }
--- a/companion/scripts/theme/applyRoles.ts
+++ b/companion/scripts/theme/applyRoles.ts
@@ -29,6 +29,7 @@ import { readFileSync, writeFileSync } from "node:fs";
 import {
   DASHBOARD_CSS_PARTS,
   DASHBOARD_PATH as DASHBOARD,
+  THEME_REGISTRY_PATH as REGISTRY_FILE,
   THEME_CSS_PARTS,
   loadBaseline,
   THEME_SOURCES,
@@ -112,11 +113,17 @@ function main() {
   // sync with the CSS from one run, because the two failure modes are both silent — a menu entry
   // with no matching block renders as the previous theme under a new name, and a block with no
   // entry is unreachable.
-  const rs = htmlSrc.indexOf(REGISTRY_BEGIN);
-  const re = htmlSrc.indexOf(REGISTRY_END);
-  if (rs < 0 || re < 0) throw new Error("theme registry markers not found in dashboard.html");
-  let outHtml =
-    htmlSrc.slice(0, rs) + renderThemeRegistry(IMPORTED_THEMES, values) + htmlSrc.slice(re + REGISTRY_END.length);
+  // The registry lives in js/dashboard-theme.js since #415, not in dashboard.html. Reading it from
+  // the wrong file threw here on every run, so the generator had been unable to run at all.
+  const registrySrc = readFileSync(REGISTRY_FILE, "utf8");
+  const rs = registrySrc.indexOf(REGISTRY_BEGIN);
+  const re = registrySrc.indexOf(REGISTRY_END);
+  if (rs < 0 || re < 0) throw new Error(`theme registry markers not found in ${REGISTRY_FILE}`);
+  let outRegistry =
+    registrySrc.slice(0, rs) +
+    renderThemeRegistry(IMPORTED_THEMES, values) +
+    registrySrc.slice(re + REGISTRY_END.length);
+  let outHtml = htmlSrc;
 
   // Call-site rewrite. Both patterns are anchored on the full 6-hex name so a partial
   // match is impossible, and an unknown name is left alone and reported rather than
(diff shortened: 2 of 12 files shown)


hasamba:
## ⚠️ Do not merge yet — I found a defect in this PR

While starting the next baseline I discovered that **the token region I edited is generated code**. `public/css/dashboard-tokens.css` says so in its own header:

> THE TOKEN REGION BELOW IS GENERATED. `npm run theme:apply` rewrites everything between the marker comments from `companion/scripts/theme/`. Do not hand-edit: the next run overwrites it.

So three of my four token changes (`--text-faint`, `--badge-warning-text`, `--text-muted`/`--tag-gray-text`) sit in a region that a regeneration will silently revert — while the a11y ledger stays at zero. That is worse than not fixing it: the gate would report clean while the violations came back.

### Why it isn't reverting today

`npm run theme:apply` is **already broken on master**, unrelated to this PR:

```
Error: theme registry markers not found in dashboard.html
    at main (companion/scripts/theme/applyRoles.ts:117)
```

The registry moved to `public/js/dashboard-theme.js` in #415, but `applyRoles.ts` still looks for it in `dashboard.html`. The generator throws before writing anything, which is the only reason my hand-edit survives.

### What the durable fix actually is

The root cause is in the generator, and it's more valuable than what this PR does:

1. **`textRampSteps()` (`themeCss.ts:201`) floors contrast at 3.0.** `--text-faint` is small body text, so AA needs **4.5**. This is why it measured 3.57:1 — it was generated to be sub-AA by design.
2. **It solves against `--bg-primary` only.** Every failure I measured was on `--bg-secondary` (`#eaeef5`), which is darker in light themes — so the solved value doesn't hold where the text actually sits. The severity solver next to it already takes an *array* of backgrounds; the text ramp should too.
3. The tier-A light literals (`--badge-warning-text`, `--text-muted`, `--tag-gray-text`) belong in `roleMap.ts`, not in the generated CSS.

Fixed there, this corrects **all 24 themes** instead of only the default light one — and axe only ever scans one theme, so the other 23 are unmeasured today.

### What is still good in this PR

Two changes are **not** in generated regions and are correct as-is:

- `public/dashboard.html` — 28 secondary/cancel buttons given a text colour, plus the 6 select labels (`for=` / `aria-label`)
- `public/js/dashboard-scope.js` — `#scopeInfo` no longer hardcodes dark-palette hex values

Recommend: hold this PR, and let me redo it as *repair the theme generator, raise the text-ramp floor to AA, regenerate*. Same outcome, but it survives the next `theme:apply` and covers every theme.

hasamba:
## Fixed — this PR is now correct and ready

Pushed `c4dac370`, which supersedes the hand-edit. **49 → 0, from the generator**, so it survives the next `theme:apply` and covers all 24 themes rather than the one axe scans.

Three real defects, all upstream of the CSS:

**1. The generator could not run.** `theme:apply` threw `theme registry markers not found in dashboard.html` — the registry moved to `js/dashboard-theme.js` in #415 and `applyRoles.ts` wasn't moved with it. Every run died before writing. That's what made a generated file look hand-editable, and it's exactly the trap the first version of this branch fell into.

**2. The text ramp was generated below AA.** `textRampSteps` floored the faintest step at **3.0:1** — the *large-text* floor — while axe found `--text-faint` on 28 nodes of ordinary small body text at 3.57:1. Now floors at 4.5 and solves against the **worst** of `--bg-primary`/`--bg-secondary`, since faint text sits on panels as often as on the page and in a light theme the panel is the darker surface. The severity solver beside it already took both backgrounds.

Same solving now applies to the built-in light/dark themes, which had been carrying legacy values through untouched — so the floors held everywhere *except* the two themes most people see.

**3. `solveForContrast` cannot raise contrast.** It interpolates between background and ink, so it only ever *dims*; asked to raise contrast it returns the ink unchanged. Correct for the ramp, silently useless as an AA guard — my first attempt at the tier-A floor looked like it worked and changed nothing. New `deepenForContrast` moves the ink toward black on light backgrounds and white on dark ones, leaving passing colours untouched.

It also bisects on the **quantised** colour: solving in float and rounding afterwards left three values at 4.47:1 against a 4.5 target — passing in the solver, failing in the browser.

### Unchanged from before (not generated, still correct)

28 secondary/cancel buttons that inherited white text onto light grey (1.36:1); six selects with no accessible name; `#scopeInfo` no longer hardcoding dark-palette hex (1.37:1 on white in light theme).

### Verification

- axe **49 → 0** across all five scopes
- theme suite **201 passed** (contrast/ordering guards read the generated blocks, so they'd fail if the generator regressed)
- full E2E **239 passed**
- typecheck, eslint, prettier, size/import/boundary ratchets all green

PR #548, opened 2026-08-10
refactor(routes): split two oversized route modules off the size ledger

hasamba:
Third of the four grandfathered baselines, **in progress**. Ledger **13 → 11 files**; two files graduated off it entirely rather than being re-recorded at a smaller number.

| File | Before | After | Result |
|---|---:|---:|---|
| `routes/caseLifecycle.ts` | 1204 | **791** | off the ledger |
| `routes/velociraptor.ts` | 888 | **751** | off the ledger |

## Why these two first

Both actively obstructed earlier work in this series. `caseLifecycle.ts` forced the atomic case-claim fix to pay for five new lines by *deleting comments elsewhere in the file*, and forced the export-snapshot fix onto a single line. A ledger that makes a security fix arrive with less documentation is working against the thing it protects.

## The seams

**`caseLifecycle.ts` → `casePush.ts`** (417 lines). Outbound pushes — DFIR-IRIS, Timesketch, MISP, Notion, ClickUp, Jira, ServiceNow, plus the status/reconnect endpoints they depend on. `caseLifecycle` is about a case's *own* lifecycle (create, archive, restore, delete, export); every push route is about a third-party API, its credentials and its failure modes. A Jira change and a case-deletion change no longer share a file.

**`velociraptor.ts` → `velociraptorMonitors.ts`** (139 lines). Everything else in that module is request/response — run a query, launch a hunt, collect from a host — a bounded exchange that finishes. A monitor is a *standing subscription*: created, polled on a timer, stopped, restarted, removed, owning background timers nothing else there touches.

Both moved **verbatim**: route bodies, status codes and log lines unchanged, only the registration point, following the `registerSeedDemoRoutes` / `registerEncryptedImportRoutes` convention already used here. `logLine` is re-declared in each new module for the same reason the originals declare it — so ~30 moved call sites stay byte-identical rather than being rewritten to `serverLogger.info`.

## A note on Prettier

Running Prettier over `velociraptor.ts` inflated it **753 → 1127**, which would have blown the very cap this change exists to clear. Both files predate the formatter. Reformatting them is a separate decision from splitting them — bundling the two would bury a 139-line move inside a 1,100-line diff. (Same trap as the earlier bounded-output fix, where a reformat had to be reverted for the identical reason.)

## Remaining

11 files still on the ledger, including the genuinely large ones: `routes/import.ts` (1766), `public/dashboard.html#inline-js` (1964), `analysis/velociraptorImport.ts` (1503), `reports/markdown.ts` (1461). Those need real decomposition rather than a clean seam, and each carries more regression risk — they are better as their own PRs than appended here.

## Test plan

- [x] `caseLifecycle` split: server + routes suites **1056 passed**
- [x] `velociraptor` split: routes suites **962 passed**
- [x] typecheck, eslint, prettier clean on both
- [x] Ledger updated via `--update` (reductions only; no `--init`, nothing re-baselined upward)
- [x] Boundary, import-cycle and a11y ratchets unmoved


hasamba:
## Third split added — ledger **13 → 10**

| File | Before | After | Result |
|---|---:|---:|---|
| `routes/caseLifecycle.ts` | 1204 | **791** | off the ledger |
| `routes/velociraptor.ts` | 888 | **751** | off the ledger |
| `routes/mcp.ts` | 845 | **715** | off the ledger |

`routes/mcp.ts` → `routes/mcpServers.ts`: the server registry (status, CRUD, discovery, tools, reconnect). The seam is **ownership of state** — everything left in `mcp.ts` *runs* something against a case and reads the registry as configuration; these routes are its only writers. `discovered` and `toolsByServer` moved **with** them rather than being shared back: they're caches invalidated by `POST /mcp/reconnect`, and leaving them behind would mean two modules mutating one cache with no owner.

## A finding worth recording: Prettier blocks the data-heavy files

I attempted `analysis/seedDemoCase.ts` (819) next and **reverted it**. It's worth explaining, because it applies to several files still on the ledger.

That file is ~390 lines of *logic* and ~390 lines of literal demo evidence. The obvious split — evidence into its own module — runs into a hard constraint:

- New files must be Prettier-clean (the format gate requires it).
- These files **predate Prettier**, and it expands dense data literals about **fourfold**.

The extracted data module measured 424 lines unformatted and **1651 formatted** — a brand-new entry on the very ledger this work is clearing. Splitting it three ways by kind (findings / IOCs / timeline) still left the timeline at 1091, and reformatting the *remainder* pushed `seedDemoCase.ts` itself from 426 to 1067 — over its own cap.

So for a data-heavy pre-Prettier file, every available split is **net negative for the ledger**. The honest move was to revert rather than trade one 819-line file for a 1067 and a 1091.

This is the same trap that forced a Prettier revert in the bounded-output fix (#541) and again in `velociraptor.ts` above. It suggests the real prerequisite is a decision about formatting these legacy files — a `.prettierignore` entry for data modules, or a one-off reformat-and-rebaseline — rather than more splitting.

## Where that leaves the remaining 10

Cleanly splittable (logic, clear seams): `routes/import.ts` (1766), `integrations/velociraptor/velociraptorApi.ts` (1086), `reports/markdown.ts` (1461).

Blocked by the Prettier expansion above: `analysis/seedDemoCase.ts` (819), and likely `memoryImport.ts` / `siemImport.ts` / `velociraptorImport.ts`, which are largely parser tables and fixtures.

Separate problem entirely: `public/dashboard.html#inline-js` (1964) and `public/js/dashboard-velo-triage.js` (815) are browser scripts, not modules.

PR #551, opened 2026-08-11
fix(tests): stop negative tests printing their own refusals onto every run

hasamba:
## The report

Every companion test run prints this, and has for months:

```
[inventory] FAIL: sections cover 2 lines but the inline script is 4. A block sits outside every banner comment and is invisible to this inventory.
```

The suspicion to check first was that `scripts/dashboard-inventory.mjs` had gone stale — #415 moved the inline JavaScript out into `public/js/*.js`, and "4 lines of inline script" is implausible for a 20k-line dashboard, so the tool was probably never repointed.

## What it actually is

**The tool is fine.** Run against the real file it passes:

```
$ node scripts/dashboard-inventory.mjs
[inventory] inline script: 1774 lines in 14 sections (1774 covered)
$ echo $?
0
```

`HTML_PATH` resolves to `public/dashboard.html` (474 KB, present), the locator finds the block, and coverage is exact. Nothing needed repointing and no banner was missing.

The "4 lines" is a **fixture built inside the test**. `dashboardInventory.test.ts` has a negative test that hands the script a deliberately broken 4-line dashboard and asserts it refuses — a guard, working. Three more like it live in `dashboardSectionSplit.test.ts`, which is where the `[split] REFUSING:` lines come from. All 27 inventory tests passed while printing FAIL.

**The defect is the output.** `execFileSync`'s default `stdio` writes the child's stderr to the *parent's* stderr as well as capturing it into the thrown error. Provoking a refusal therefore prints it. Verified directly:

```
parent-saw: LEAK-DEFAULT          # {encoding:"utf8"}
                                  # {stdio:["ignore","pipe","pipe"]} — nothing
```

`error.stderr` is populated in both cases, so nothing depended on the echo.

That is worth fixing rather than shrugging at: a green run that prints `FAIL` teaches people to read past the word, so the day the coverage check fires against the real dashboard it scrolls by as more of the usual noise. A check nobody will believe has stopped being a check.

## The change

- **`tests/helpers/runScript.ts`** (new) owns the spawn, with `stdio` set so the child's stderr lands only in the result. `runScript` / `runScriptJson` / `runScriptExpectingFailure`.
- **`dashboardInventory.test.ts`**, **`dashboardSectionSplit.test.ts`**: all 21 spawn sites go through it. The four refusal cases lose their try/catch — `runScriptExpectingFailure` throws when the script *succeeds*, so a guard that has stopped guarding fails the test instead of silently skipping the assertions that lived in a catch block.
- **`childStderr.test.ts`** (new) pins what nothing was watching:
  - no test spawns a script except through the helper (the leak was four copies of one options object; a fifth is how it comes back);
  - the refusal text is *still* available to assert on — a helper that swallowed stderr outright would satisfy the check above and quietly gut every refusal assertion in the suite;
  - a script that stops refusing fails loudly.

Both sibling `[split]` leaks are fixed here too: same root cause, same options object, same fix.

## End state

`npx vitest run tests/architecture/` — 63 passed, and grepping its output for `[inventory]` / `[split]` / `REFUSING` / `FAIL:` returns nothing. The coverage check now passes silently and would genuinely fail loudly.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean
- [x] `npm run format:check` — clean
- [x] `npx vitest run` — **582 files, 9165 tests, all passing**
- [x] `node scripts/dashboard-inventory.mjs` — exits 0, 1774/1774 covered
- [x] Full architecture run greps clean for leaked child stderr

### DIFF
--- a/companion/tests/architecture/childStderr.test.ts
+++ b/companion/tests/architecture/childStderr.test.ts
@@ -24,6 +24,39 @@ import { runScriptExpectingFailure } from "../helpers/runScript.js";
 
 const TESTS = new URL("../", import.meta.url);
 
+/**
+ * Line numbers of spawns in `source` that would echo the child's stderr onto this process's.
+ *
+ * The rule is the invariant itself — every spawn passes an explicit `stdio` that pipes stderr —
+ * rather than "call the helper", which is only one way of satisfying it. Enforcing the proxy would
+ * fail a spawn that is already correct, and a guard that fires on correct code gets relaxed by
+ * whoever hits it next, usually by deleting it.
+ *
+ * Reads the argument list by matching parens rather than by line, because the four call sites that
+ * caused this all wrapped their options object onto its own lines. A paren inside a string literal
+ * would confuse the balance; none of these calls has one, and the failure mode is a false positive
+ * naming a real line, which is the safe direction.
+ */
+export function unpipedSpawns(source: string): number[] {
+  const found: number[] = [];
+  const call = /\b(?:execFileSync|execSync|spawnSync)\s*\(/g;
+  for (let m = call.exec(source); m; m = call.exec(source)) {
+    const open = m.index + m[0].length - 1;
+    let depth = 0;
+    let i = open;
+    for (; i < source.length; i++) {
+      if (source[i] === "(") depth++;
+      else if (source[i] === ")" && --depth === 0) break;
+    }
+    const args = source.slice(open + 1, i);
+    // `stdio: "inherit"` is the loud way to do exactly what the default does quietly.
+    if (!/\bstdio\s*:/.test(args) || /["']inherit["']/.test(args)) {
+      found.push(source.slice(0, m.index).split("\n").length);
+    }
+  }
+  return found;
+}
+
 /** Every .ts file under tests/, as a path relative to tests/. */
 async function testSources(dir: URL, prefix = ""): Promise<string[]> {
   const out: string[] = [];
@@ -55,11 +88,10 @@ describe("tests do not leak a child process's stderr into the run", () => {
 
   afterAll(() => rmSync(dir, { recursive: true, force: true }));
 
-  it("spawns scripts only through the stderr-capturing helper", async () => {
+  it("gives every spawn in the suite an explicit stdio that pipes stderr", async () => {
     // Structural, not behavioural, and deliberately so. The leak is invisible from inside the
     // process that leaks — the parent's stderr belongs to the test runner — so the cheap thing a
-    // test can check is that no call site sets the leak up. One helper owns the `stdio` option;
-    // any other spawn is a copy of the options object waiting to drop it again.
+    // test can check is that no call site sets the leak up.
     const files = await testSources(TESTS);
     expect(files.length, "found no test sources — the walk is looking in the wrong place").toBeGreaterThan(
       10,
@@ -67,21 +99,37 @@ describe("tests do not leak a child process's stderr into the run", () => {
 
     const offenders: string[] = [];
     for (const file of files) {
-      if (file === "helpers/runScript.ts") continue;
+      // This file is the one place the offending shape appears as DATA — the fixture in the test
+      // below is a string of deliberately bad spawns, and scanning it reports itself. Its own two
+      // spawns go through the helper, so nothing here is exempted from the rule in practice.
+      if (file === "architecture/childStderr.test.ts") continue;
       const source = await readFile(new URL(file, TESTS), "utf8");
-      for (const [i, line] of source.split("\n").entries()) {
-        // The call form, not the bare name: the helper is named in prose in more than one comment,
-        // and a mention is not a spawn.
-        if (/\b(?:execFileSync|spawnSync)\s*\(/.test(line)) offenders.push(`${file}:${i + 1}`);
-      }
+      offenders.push(...unpipedSpawns(source).map((line) => `${file}:${line}`));
     }
 
     expect(
       offenders,
-      "spawn scripts via tests/helpers/runScript.ts — a bare execFileSync echoes the child's " +
-        "stderr onto the run's own, which is how four refusal messages came to print on every " +
-        "green run",
+      "these spawns would echo the child's stderr onto the run's own, which is how four refusal " +
+        "messages came to print on every green run. Use tests/helpers/runScript.ts, or pass " +
+        '`stdio: ["ignore", "pipe", "pipe"]` yourself',
     ).toEqual([]);
+  });
+
+  it("flags a spawn that omits stdio, and only that one", () => {
+    // The detector, watched working. Scanning real files reports "[]" both when every spawn is
+    // correct AND when the scan has quietly stopped recognising spawns at all — two cases a
+    // file-walking assertion cannot tell apart on its own.
+    const source = [
+      'const a = execFileSync(node, [s], { encoding: "utf8" });', // 1: no stdio
+      "const b = execFileSync(node, [s], {",
+      '  encoding: "utf8",', // options wrapped across lines, as all four leaks were
+      "});", // reported at line 2, where the call starts
+      'const c = execFileSync(node, [s], { encoding: "utf8", stdio: ["ignore", "pipe", "pipe"] });',
+      'const d = spawnSync(node, [s], { stdio: "inherit" });', // 6: loudly does the default
+      "const e = runScript(SCRIPT, [s]);", // not a spawn at all
+    ].join("\n");
+
+    expect(unpipedSpawns(source)).toEqual([1, 2, 6]);
   });
 
   it("still hands the refusal text to the test that asked for it", () => {
--- a/companion/tests/architecture/childStderr.test.ts
+++ b/companion/tests/architecture/childStderr.test.ts
@@ -19,42 +19,85 @@ import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { readFile, readdir } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
+import ts from "typescript";
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
 import { runScriptExpectingFailure } from "../helpers/runScript.js";
 
 const TESTS = new URL("../", import.meta.url);
 
 /**
- * Line numbers of spawns in `source` that would echo the child's stderr onto this process's.
+ * Which stdio slot a spawn gives its child's STDERR, or null when that cannot be determined.
  *
- * The rule is the invariant itself — every spawn passes an explicit `stdio` that pipes stderr —
- * rather than "call the helper", which is only one way of satisfying it. Enforcing the proxy would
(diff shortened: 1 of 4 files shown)


hasamba:
## Revised after review: the guard checked a proxy, not the invariant

`childStderr.test.ts` originally required every test spawn to go through `tests/helpers/runScript.ts`. That is a stand-in for the real property, and it **fails code that already satisfies it** — a spawn passing its own `stdio: ["ignore", "pipe", "pipe"]` cannot leak, and was still reported.

That made this PR and #552 mutually unshippable. #552 adds one correctly-piped `execFileSync` in `moduleMap.test.ts`, so whichever landed second would have gone red on a rule it was already obeying. I had noted it as a merge-order chore; that was the wrong call — a guard that fires on correct code gets relaxed by whoever hits it next, usually by deleting it.

`unpipedSpawns()` now reports spawns that **omit `stdio` or inherit it**, reading the argument list by matching parens rather than by line — all four original leaks wrapped their options object across lines, so a line-based scan would have missed them if they had been written slightly differently.

The detector is unit-tested against a fixture of deliberately bad spawns:

```
expect(unpipedSpawns(source)).toEqual([1, 2, 6]);   // no-stdio, multiline no-stdio, "inherit"
                                                    // (piped call and a helper call not flagged)
```

Scanning real files reports `[]` both when every spawn is correct *and* when the scan has quietly stopped recognising spawns at all — a file-walking assertion cannot tell those apart on its own. That fixture is also why the scan skips this one file: it is the only place the offending shape appears as data, and this file's own spawns go through the helper.

### Verified by actually merging both branches

```
Test Files  582 passed (582)
     Tests  9168 passed (9168)
```

…with no leaked child stderr, and `typecheck` / `lint` / `format:check` / `check:boundaries` all clean on the merged tree. **The two PRs now land in either order.**

This branch alone: 582 files / 9166 tests passing, all gates clean.

hasamba:
## Second revision: the detector was checking the wrong thing

Review caught that the previous version did not actually enforce that stderr is piped — it reported a spawn as safe if it passed **any** `stdio` key not containing the word `"inherit"`. Measured against Node:

| form | child's stderr |
|---|---|
| `stdio: "pipe"` | captured |
| `stdio: ["ignore", "pipe", "pipe"]` | captured |
| `stdio: ["ignore", "pipe", "ignore"]` | captured |
| `stdio: ["ignore", "pipe"]` | captured — Node fills the missing slot with `"pipe"` |
| *(no stdio)* | **LEAKS** |
| `stdio: "inherit"` | **LEAKS** |
| `stdio: ["ignore", "pipe", "inherit"]` | **LEAKS** |
| `stdio: ["ignore", "pipe", 2]` | **LEAKS** |

The last row is the one that mattered: **file descriptor 2 *is* the parent's stderr**, and there is no keyword in that text to search for. The old check passed it.

`unpipedSpawns()` now reads the `stdio` value and asks what happens to stderr specifically — a bare string applies to all three streams, an array is indexed at slot 2, a short array takes Node's `"pipe"` default. A value that is not a literal (identifier, spread, ternary) is **reported rather than assumed safe**; that conservatism applies to `runScript` itself, whose stdio triple is now written out literally instead of spread from a constant.

### The table is asserted by running it, not by reading the docs

Two versions of this guard now have looked correct and not been. So each row is executed: a child spawns a grandchild that writes to stderr, we capture the child's stderr, and the observed leak is compared against what the detector says about the same option text.

```
✓ agrees with what each stdio form actually does to a real child's stderr
```

Non-vacuous by construction — the detector flags 4 of the 8 forms and the assertion is equality with the observed leak, so it can only pass if exactly those 4 leaked. It also pins Node's short-array default: if that ever changes, this fails instead of the guard quietly going wrong.

### Still ships with #552 in either order

Re-merged and re-run after the change: **582 files, 9169 tests**, `typecheck` / `lint` / `format:check` / `check:boundaries` clean, no leaked child stderr.

This branch alone: 582 files / 9167 tests passing.

hasamba:
## Third revision: spreads defeated the positional read

Review caught a false negative I introduced with the fix for the last one. `stdio: [...STDIO]` has **fewer than three top-level elements**, so my "short array takes Node's `pipe` default" rule returned `"pipe"` — while the constant could expand to three entries ending in `"inherit"`. `["ignore", ...rest]` is the same hole with one slot known.

The form is not hypothetical: **the first draft of `runScript.ts` in this very PR used `stdio: [...STDIO]`.** The detector I wrote would have blessed it.

Watched failing before fixing:

```
× flags every spawn whose stderr reaches this process, and no others
  → expected [ 1, 2, 6, 7, 8, 10 ] to deeply equal [ 1, 2, 6, 7, 8, 10, 14, 15 ]
      -   14,     // stdio: [...STDIO]
      -   15,     // stdio: ["ignore", ...rest]
```

Any spread in the array now returns "unreadable", joining the values that already did — identifiers, ternaries, stream objects. I also anchored the bare-string form so `stdio: "pipe" + x` is not read as if the literal were the whole value.

### The asymmetry is deliberate

Over-reporting costs someone one explicit `stdio`. Under-reporting cost this repo months of `FAIL` scrolling past on every green run. So unreadable ⇒ reported.

That is also why the behavioural cross-check covers only statically-readable forms: running `[...STDIO]` would compare a deliberate false positive against a clean result and fail. Those forms are pinned in the fixture instead, and the comment now says so.

### State

Three versions of this guard have now looked correct and been wrong — which is the same defect as the bug the PR fixes, and the reason the leak/no-leak table is asserted by execution rather than by reading the docs.

Re-merged with #552 and re-run after the change: **582 files, 9169 tests**, `typecheck` / `lint` / `format:check` / `check:boundaries` clean, no leaked child stderr. Either merge order still works.

This branch alone: 582 files / 9167 tests passing.

hasamba:
## Fourth revision: comments broke the scan in both directions

Review caught that spread detection missed **commented** spreads. It did, and the same root cause was also producing a false *positive*. Both watched failing first:

```
× flags every spawn whose stderr reaches this process, and no others
  → expected [ 1, 2, 6, 7, 8, 10, 14, 15, 23 ]
    to deeply equal [ 1, 2, 6, 7, 8, 10, 14, 15, 16, 17 ]
```

- **16, 17 missing (false negatives).** `stdio: [/* keep */ ...STDIO]` — and its multi-line form with a `//` comment above the spread — do not have an element *starting* with `...`, so the spread check passed them, the array counted as short, and it took Node's `"pipe"` default.
- **23 present (false positive).** A call inside a comment (`// execFileSync(node, [s], {})`) was parsed as a real spawn. This suite is heavily commented, so that direction was live too.

## Fix

Comments are blanked before any parsing, replaced with spaces so **every offset — and therefore every reported line number — still lines up** with the original source. The masker is string-aware, so `//` inside a URL is not a comment start; and the paren-matcher and element-splitter now step over string literals, which also retires a caveat I had written into the doc comment rather than fixed.

## Where this leaves it

Four versions of this guard have now looked correct and been wrong — the same defect as the bug the PR fixes. What has actually caught them: the leak/no-leak table is asserted **by execution** (child spawns grandchild, we capture the child's stderr, compare against the detector's verdict on the same text), and every structural case is pinned in a fixture with the expected line numbers, so a regression shows up as a diff of line numbers rather than as silence.

Re-merged with #552 and re-run: **582 files, 9169 tests**, `typecheck` / `lint` / `format:check` / `check:boundaries` clean, no leaked child stderr. Either merge order still works. This branch alone: 582 files / 9167 tests.

hasamba:
## Fifth revision: the scanner read text that looked like code

Review caught that inert calls inside string literals were still reported. They were — and that false positive was the reason this file **exempted itself from its own rule**, which I had written up as a design note rather than treating as the bug it was.

Watched failing first:

```
× flags every spawn whose stderr reaches this process, and no others
  → expected [ 1, 2, 6, 7, 8, 10, 16, 17, 24 ] to deeply equal [ 1, 2, 6, 7, 8, 10, 14, 15, 16, 17 ]
```

Line 24 is `describe("execFileSync(node, [s], {})", …)` — a call quoted as data.

## Fix

String literals are now ranges the scanner steps over rather than parses, alongside the comment masking from the previous round. Together the scanner reads code and not text that resembles it, and it was wrong in three ways at once before:

| input | old behaviour |
|---|---|
| comment in front of a spread | element didn't start with `...` → **missed** |
| call named in prose | parsed as real → **false positive** |
| call quoted as data | parsed as real → **false positive** |

**The self-exemption is gone.** `childStderr.test.ts` is now scanned like every other file and reports clean. An exemption is a hole the size of whatever else is in the file, and I should not have shipped one.

## State

Re-merged with #552 and re-run: **582 files, 9169 tests**, `typecheck` / `lint` / `format:check` / `check:boundaries` clean, no leaked child stderr. Either merge order works. This branch alone: 582 files / 9167 tests.

Five rounds of review have now found five defects in this guard, each one the same shape as the bug the PR fixes — a check that reads as working. Every one was found by review rather than by my own testing, so I would not treat this file as settled without another pass.

hasamba:
## Sixth revision: stopped scanning, started parsing

Review caught that filtering out string literals introduced a **known false negative** — a real call inside a template interpolation. I had documented that as acceptable ("nobody writes one there") one round after arguing that under-reporting is the one direction that matters. That was rationalising, not reasoning.

Pinned it first:

```
× flags every spawn whose stderr reaches this process, and no others
  → expected [ …10 ] to deeply equal [ …, 25 ]     // `${execFileSync(node, [s], { encoding: "utf8" })}`
```

## The real problem was the approach

Six rounds, six defects. Only the **first** was about `stdio`. The rest were all lexical:

| round | defect |
|---|---|
| 1 | proxy rule ("use the helper") failed correct code |
| 2 | any `stdio` key without `"inherit"` counted safe — `[..., 2]` passed |
| 3 | `[...STDIO]` counted as a short array, took the `"pipe"` default |
| 4 | a comment in front of a spread hid the spread |
| 5 | a call in prose / quoted as data parsed as real |
| 6 | filtering strings hid a real call in a template interpolation |

Each fix was correct and the next edge case was already queued. Patching a hand-rolled lexer was the wrong shape of solution.

**It now parses.** `typescript` is already a dependency here, and `scripts/dashboard-inventory.mjs` reaches for it for exactly this reason — its header says to read the AST rather than the text, having been burned by grep. I should have followed the precedent already in this repo. The whole lexical class is gone rather than narrowed, and ~100 lines of scanner went with it.

What's left is the actual question: `stderrSlot()` walks the options object of a spawn call and reports what reaches stderr. Anything that hides the answer — a spread in the options or the array, a non-literal value — is reported rather than assumed safe.

## Verification unchanged in kind, stronger in coverage

Every row of the leak table is still asserted **by execution** (child spawns grandchild, we capture the child's stderr, compare with the detector's verdict on the same text). The fixture now carries every lexical case from the table above, so none can come back quietly.

Re-merged with #552: **582 files, 9169 tests**, all gates clean, no leaked child stderr. Either merge order works.

Worth stating plainly: all six defects were found by review, none by my own testing, and the guard is optional scaffolding around a one-line `stdio` fix that has been correct since the first commit. If it keeps costing more than it protects, dropping `childStderr.test.ts` and keeping the `runScript.ts` helper is a reasonable call.

PR #600, opened 2026-08-20
fix(import): read a script block whose raw event sits under `_Event`

hasamba:
Follow-up to #599, found by running that change against two real collections.

## The gap

A 4104 row reaches the Velociraptor importer in **three** shapes, not two:

1. native parsed evtx — top-level `System` + `EventData`
2. wrapped — the same under `Event`
3. **Velociraptor-decorated — the same under `_Event`**, alongside rendered `Details`/`Title` columns

`Windows.Hayabusa.Rules` emits shape 3, and the row lands in *this* importer rather than the Hayabusa one because it carries `_Source`, which the Velociraptor detector claims.

#599's fragment reader knew shapes 1 and 2 only. A script block split across several events in shape 3 stayed one alert per fragment, with nothing to signal the miss. `rowMessage()` already read `_Event`, so the shape was known to the file but not to the reader.

## The fix

Teach `veloEventData` / `veloEventId` / `veloHost` the `_Event` wrapper, and **resolve** the write-back container instead of assuming it — guessing wrong left the real `EventData` untouched *and* invented a sibling key the row never had.

## Measured on real collections

A flow export carrying DetectRaptor + Chainsaw + Hayabusa together:

| | events |
|---|---|
| before (`57f1bf99`) | 1410 |
| after | **1072** |

The 448 Hayabusa rows carrying a `ScriptBlockId` all use the `_Event` shape and were untouched until now. Their 374 fragments sit in 63 blocks, and **every fragment is a distinct position** — nothing is over-merged. Largest merged alert: 33 source records, 20 reassembled parts.

A second collection whose 4104 rows are all single-part (`MessageTotal = 1`) is **unchanged at 855 events**, which is the correct no-op.

The duplicate-position dedupe from #599 is doing real work on this data too: DetectRaptor has 136 records across 129 distinct positions, Chainsaw 77 across 43.

## Verification

- Full suite **9870 passing** (640 files).
- `typecheck`, `lint`, `check:size`, `check:imports`, `check:boundaries`, `format:check` all clean.
- New tests are built from the **real row shape**, not a synthetic guess, and cover both the split block and the single-part block that must stay untouched.

## Known limitation, unchanged by this PR

A *standalone* `hayabusa` export — one whose rows carry only a rendered `Details` string with no `_Event` node — still cannot be consolidated. Such rows have no `ScriptBlockId`, `MessageNumber` or `MessageTotal` anywhere, so there is no block identity to join on. That is a collection-profile question, not a parser one.

### DIFF
--- a/companion/src/analysis/scriptBlockFragments.ts
+++ b/companion/src/analysis/scriptBlockFragments.ts
@@ -171,12 +171,28 @@ function reassembleByBlock(
 
 // ───────────────────────────── Velociraptor / DetectRaptor rows ─────────────────────────────
 
-// A 4104 row reaches the importer either as a native parsed-evtx row (`System` + `EventData`,
-// sometimes wrapped in `Event`) or — after an Elasticsearch round-trip — as dotted keys that
-// veloRowNormalize has already un-flattened back to that nested shape. Both are read here; a row
-// in neither shape simply yields no fragment and is left alone.
+// A 4104 row reaches this importer in one of THREE shapes, all of which must be read:
+//
+//   1. native parsed evtx      — top-level `System` + `EventData`
+//   2. wrapped                 — the same under `Event`
+//   3. Velociraptor-decorated  — the same under `_Event`, alongside rendered `Details`/`Title`
+//      columns. `Windows.Hayabusa.Rules` emits this, and the row reaches HERE rather than the
+//      Hayabusa importer because it carries `_Source`, which the Velociraptor detector claims.
+//
+// An Elasticsearch round-trip arrives as dotted keys that veloRowNormalize has already un-flattened
+// back into shape 1. A row in none of these yields no fragment and is left untouched.
+//
+// Shape 3 was missed at first and is why this list is spelled out: rowMessage() in velociraptorImport
+// already read `_Event`, so the shape was known to the codebase but not to this reader, and a split
+// block collected that way stayed one alert per fragment with nothing to signal the miss.
+const EVENT_WRAPPERS = ["Event", "_Event"] as const;
+
 function veloEventData(row: Row): Row | null {
-  const ed = getCI(row, "EventData") ?? getPath(row, "Event.EventData");
+  let ed = getCI(row, "EventData");
+  for (const w of EVENT_WRAPPERS) {
+    if (isObject(ed)) break;
+    ed = getPath(row, `${w}.EventData`);
+  }
   return isObject(ed) ? ed : null;
 }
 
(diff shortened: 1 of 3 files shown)


hasamba:
## Review round — both findings verified and fixed

Both were real. Verified by execution before changing anything.

### P1 — a verdict was being dropped (confirmed)

Two different rules over one `_Event` block produced **one** event, and the second rule's verdict was gone from the case:

```
events: 1        (2 = verdicts kept, 1 = a verdict was LOST)
'UAC Bypass Attempt' still present: false
```

Cause is as described: a `_Event` Hayabusa row keeps its rule name in a top-level `Title` that `classify()` reads as generic, so `mapGeneric` puts it in neither the description nor the aggregation key. The differing fragment text was the only thing holding two rules apart, and this PR removed it.

**Fix:** the reassembly group is now scoped by rule identity as well as host and `ScriptBlockId`.

**This withdraws a property #599 promised** — "every alert carries the whole script". An alert now joins the parts *its own rule* matched, and the `N of M` note discloses that the block had more. A shorter excerpt is a smaller harm than a missing verdict. The test asserting the old contract is updated to state the new one rather than deleted.

Cost on real data is negligible: **1072 → 1076 events**, because those Hayabusa rows all share one rule title. Against master the net is still **1410 → 1076**.

### P2 — stale nested message (confirmed)

An `_Event` row with no top-level `Message`/`Details` did not consolidate at all — 2 events where 1 was expected — because `rowMessage()` falls back to `_Event.Message`, which kept the fragment text and so kept the fingerprints distinct.

**Fix:** every rendered copy the row carries is rewritten, not just the first found. `Message` and `Details` can both be present, and a copy left holding one slice contradicts the others.

### One thing found while fixing P2

The fallback rewrite path left the `(N of M)` header uncorrected, so a joined message still read `(1 of 2)` above the full text — the same "reads as truncation" problem the structured path already fixed. Now corrected on both paths. It edits only the prefix *before* the fragment's text begins, so a `(1 of 2)` occurring inside the script itself is never rewritten.

### Still open, deliberately out of scope

`mapGeneric` never renders the Hayabusa rule title into the description at all, so these alerts read `Velociraptor [Windows.Hayabusa.Rules]: ScriptBlock: …` with no rule name. That predates both PRs — the two alerts stay separate now, but an analyst cannot tell from the row which rule fired. Fixing it means teaching `classify()`/`mapGeneric` to treat these as verdict rows, which affects every artifact on that path and belongs in its own change.

Full suite **9873 passing**; typecheck, lint, size, imports, boundaries and format all clean.

Issue #892, opened 2026-09-08
[defang] Interactive HTML report and standalone presentation deck expose live indicators as clickable text


hasamba:
**File:** `companion/src/reports/interactiveHtml.ts`, `companion/src/analysis/presentation.ts`
**Commits reviewed:** `ea0e375`–`25773149` (24h)

## Problem

The new `defang.ts` pass defangs indicators in the Markdown output, the HTML report, and the DOCX report — but the same indicators appear unprotected in two other human-readable exports:

1. **`renderInteractiveHtmlReport`** (`interactiveHtml.ts`) — a self-contained HTML file designed to be emailed. IOC values are rendered as plain DOM text (no `<a href>`), but the file IS the deliverable and its entire purpose is to travel to recipients outside the analyst's org. Live IPs/URLs/domains in event descriptions and IOC values are visible as plain text with no CSP.

2. **`renderStandalonePresentation`** (`presentationExport.ts`) — the offline-standalone presentation deck. Similarly embeds IOC values and event text as JSON-in-HTML, then renders them as DOM text.

Neither uses `marked` with an autolink tokenizer, so there is no active hyperlink risk. However, for the interactive HTML report the primary threat model is emailing a portable artifact to recipients who may click out of curiosity.

## Expected fix

Apply `defangIndicators()` to the event description and IOC value text in `renderInteractiveHtmlReport` before rendering — same pattern as `reportContents.ts` / `html.ts` / `docx.ts`. For the presentation deck, apply it to the `PresentationIoc.value` field before embedding.

## Suggested implementation

```ts
// In interactiveHtml.ts, before building text nodes for IOC values/descriptions:
import { caseDomains, defangIndicators } from "./defang.js";
const safe = defangIndicators(rawText, caseDomains(investigationState));
```

## Commit range
`ea0e375..25773149` (4 commits by @hasamba)

hasamba:
## Review: the gap is real, but three details here are wrong and the suggested fix is in the wrong place

### Confirmed

Neither export defangs:

- `interactiveHtml.ts` has no `defang.js` import; `toTimelineRow` (line 107) and `toFindingCard` (line 121) copy `description` straight into the embedded JSON, which the inline script renders as DOM text (lines 315, 358).
- `presentation.ts:152` `toPresentationIoc` copies `ioc.value` verbatim, and `presentationExport.ts` embeds the whole deck as `window.__DECK__`.

So live indicators do travel in both artifacts. Worth fixing.

[...]

Issue #878, opened 2026-09-06
MITRE ATT&CK panel renders "—" in all four profiles that include it


hasamba:
`#sec-mitre` renders a single em-dash in every built-in profile that includes it, while the case holds nine techniques and three sibling surfaces display them correctly.

### Affected profiles

**Analyst, Report, Deep-Dive, Hunt Prep** — all four.

### Steps to reproduce

1. Import anything that carries MITRE techniques (Suricata `eve.json` and a Linux shell history both do).
2. Switch to the Analyst profile and look at **MITRE ATT&CK**.

### Expected

The case's techniques listed, as Kill Chain and the ATT&CK Navigator export both manage.

### Actual

```
#sec-mitre  innerText  : "⠿ ▾ MITRE ATT&CK —"
#sec-mitre  innerHTML  : 161 bytes, content div contains "—"
            display    : block   (expanded, not collapsed)
```

### The data is present

Three other surfaces read it without trouble:

```
GET /cases/<case>/attack-mitigations  ->  caseTechniqueCount: 9
Kill Chain panel                      ->  Defense Evasion 1 · Credential Access 2
                                          Lateral Movement 2 · Collection 1 · C2 2
Export → ATT&CK Navigator layer       ->  13 techniques, enterprise-attack, scored
Forensic timeline rows                ->  "MITRE: T1003.003", "MITRE: T1021.002, T1570"
```

### Why it matters

Worst in the **Report** profile, where the technique list is what a reader expects to lift straight into a deliverable. Kill Chain is derived from the same tags with no AI and works, so this is not an AI-off symptom.

---
Found in an end-to-end UI QA pass of 0.36.0.

PR #895, opened 2026-09-08
fix(mitre): derive event-carried techniques at projection instead of persisting them

### LINKED ISSUE #893: [stateMerge] mergeDelta's full-timeline MITRE union defeats the synthesis scope/false-positive filter
**File:** `companion/src/analysis/stateMerge.ts:355`
**Commits reviewed:** `ea0e375`–`25773149` (24h)

## Problem

The new `unionEventTechniques` call in `stateMerge.ts` passes `correlated` — the full forensic timeline after all processing (`correlateEvents`, sorting, year-clamping, archive-linking):

```ts
// stateMerge.ts:355 — NEW call site (deterministic importers)
const mitreWithEvents = unionEventTechniques(mitreTechniques, correlated);
```

The existing call site in `synthesisMerge.ts:136` passes `scopedEvents` — only the events synthesis actually considered (time-window-filtered, false-positive-filtered):

```ts
// synthesisMerge.ts:136 — EXISTING call site (AI synthesis)
next = { ...next, mitreTechniques: unionEventTechniques(next.mitreTechniques, scopedEvents) };
```

The doc comment that moved with `unionEventTechniques` says:

> "Operates over the SAME scope/legitimate-filtered events synthesis saw, so it never reintroduces out-of-scope techniques."

This is **not true** for the new `stateMerge` call — `correlated` is not scope-filtered. While in practice the risk is low (deterministic importers don't introduce new technique IDs out-of-band, and `correlated` only includes events already in the state), the comment is misleading and the divergence between the two call sites is undocumented.

## Suggested fix

Either:
1. **Document the difference** — add a comment at the `stateMerge` call site explaining why the full timeline is intentionally used there (deterministic imports don't add out-of-scope techniques, and the MITRE panel should reflect all observed techniques).
2. **Scope it** — filter `correlated` to only events within the current scope window before passing to `unionEventTechniques`, making both call sites consistent.

Option 1 is preferred since the use case for a deterministic-import-only case does want full-timeline MITRE collection.

## Commit range
`ea0e375..25773149` (4 commits by @hasamba)

hasamba:
Rewritten. This PR is now #893 only — #892 was split out and merged as #896 — and it replaces the persisted-provenance approach with the thing that approach was working around.

## The problem

#878 fixed an empty MITRE panel by unioning the techniques a deterministic importer puts on its **events** into the case aggregate during the merge.

That aggregate is **persisted**. Scope and the false-positive filter drop events at **projection**, never from state. So once a technique was in the table there was no way back out: dismissing the only event that carried it changed nothing, and the next import re-added it anyway.

## Why the previous approach was abandoned

Nineteen review rounds went into chasing that with provenance — an exclusion filter, a merge chokepoint, per-row origins, frozen ids, cluster spans, dated vs undated, parseability, sub-millisecond ordering.

Every one of those was a genuine defect. Every one existed for the same reason: a stored aggregate has to independently reconstruct *which event a technique came from and when*, through `correlateEvents`, whose entire job is to collapse exactly that. `TechniqueOrigin` had become a second model of the timeline — identity, dating, ranges, precision — that had to be kept in step with the first at every touch point.

## What this does instead

Derive it after the filters, from the events that survive them:

```ts
export function withEventTechniques(state: InvestigationState): InvestigationState {
  return { ...state, mitreTechniques: unionEventTechniques(state.mitreTechniques, state.forensicTimeline) };
}
```

The table becomes a function of the timeline it is displayed beside, so it cannot disagree with it. A dismissal takes effect immediately, un-marking restores the technique with no merge, and nothing is written so nothing can go stale.

- **Server:** applied at `ReportWriter.loadFilteredState`, last, after both filters — so every report and export gets it.
- **Client:** mirrored in `dashboard-render.js` over `ft`, the scope-projected timeline minus dismissed events, which that view already computes for the timeline itself.
- **`mergeDelta` still persists** what a model asserts at a delta's **top level** — a conclusion about the case rather than a restatement of an event tag. That is a claim worth keeping.

## Removed

`techniqueScope.ts`, `TechniqueOrigin` and its `cluster`/`endTimestamp`/`fromEvents` flags, `techniqueContributors`, `techniqueEligibilityRefs`, `techniqueSupport`, `techniqueEventFilter`, `techniquesAreConclusions`, the second-opinion promotion, the pipeline and MCP filter wiring, and the merge-time union. `correlate.ts`, `stateTypes.ts`, `pipeline.ts`, `synthesisMerge.ts` and `secondOpinion.ts` are back to their master versions.

Net: **+242 / −44** across 9 files, against roughly 1,700 insertions for the version it replaces.

## Tests

The #878 merge-time tests move to the projection layer, joined by the three cases nineteen rounds never got right:

- a dismissal drops the technique **at once**;
- **un-marking brings it back**, with no merge in between;
- a scope window excludes a technique whose only event it excludes.

Plus unit tests for the derivation: an import-only case is populated (#878), collected rows are named, an asserted technique keeps its name and finding links rather than being restated, and the pass is idempotent and non-mutating.

Full suite **12,046 passing (740 files)**; `typecheck`, `lint`, `check:size`, `check:boundaries`, `check:imports` green.

Closes #893

https://claude.ai/code/session_016mXaxwsRxd7tffg7PDuzu3

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -219,7 +219,7 @@ established, and it works the same way.
 The graph is built the same way `check-imports.mjs` builds it: a regex over relative `.js`
 specifiers, because the companion imports its own modules exclusively that way. No resolver needed.
 
-For context: **1,981 of the 2,020 cross-domain file dependencies already comply.** The map is mostly
+For context: **1,982 of the 2,021 cross-domain file dependencies already comply.** The map is mostly
 a description of how this codebase is already written, which is the only kind of rule people follow.
 Both figures come from `npm run check:boundaries -- --json`, which counts them in the same pass that
 finds the violations, and a test asserts this sentence against it. The pair read 1,275 of 1,323 long
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -219,7 +219,7 @@ established, and it works the same way.
 The graph is built the same way `check-imports.mjs` builds it: a regex over relative `.js`
 specifiers, because the companion imports its own modules exclusively that way. No resolver needed.
 
-For context: **1,982 of the 2,021 cross-domain file dependencies already comply.** The map is mostly
+For context: **1,981 of the 2,020 cross-domain file dependencies already comply.** The map is mostly
 a description of how this codebase is already written, which is the only kind of rule people follow.
 Both figures come from `npm run check:boundaries -- --json`, which counts them in the same pass that
 finds the violations, and a test asserts this sentence against it. The pair read 1,275 of 1,323 long
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -219,7 +219,7 @@ established, and it works the same way.
 The graph is built the same way `check-imports.mjs` builds it: a regex over relative `.js`
 specifiers, because the companion imports its own modules exclusively that way. No resolver needed.
 
-For context: **1,980 of the 2,019 cross-domain file dependencies already comply.** The map is mostly
+For context: **1,982 of the 2,021 cross-domain file dependencies already comply.** The map is mostly
 a description of how this codebase is already written, which is the only kind of rule people follow.
 Both figures come from `npm run check:boundaries -- --json`, which counts them in the same pass that
 finds the violations, and a test asserts this sentence against it. The pair read 1,275 of 1,323 long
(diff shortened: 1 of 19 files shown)


hasamba:
**#892 has been split out into #896** and is no longer part of this PR's scope on the review side — the defang work was review-clean throughout and there was no reason for it to wait on the MITRE aggregate work here.

The commits remain in this branch's history (they were interleaved), so if #896 lands first this branch should be rebased on master and the duplicate dropped. If this one lands first, #896 becomes a no-op and can be closed.

This PR is now about #893 only.

Issue #930, opened 2026-09-10
Extend attack-outcome correlation and recurrence detection from IR-200 review


hasamba:
## Purpose and scope

Extract additional, evidence-led hunting and detection recommendations from the supplied IR-200 HTML training material. This is a proposal for analyst review, not approval to implement changes or perform endpoint response actions.

Reviewed against the current checkout at `f38a0958` and open recommendations #908 and #909. The source's relevant areas are incident detection, attack-chain reconstruction, eradication verification, and lessons learned. Lab commands, identities, infrastructure, and course screenshots are intentionally excluded. Source references below are section headings because the HTML has no stable page numbers.

The most valuable additions are specific correlations and explicit attack outcomes, rather than more tool-name matches. Existing email import, failed-logon grading, Kerberos heuristics, temporal grouping, evidence graphs, archive/upload correlation, false-positive handling, and response checklists should be extended where appropriate.

## Recommended first tranche

Start with **1, 3, 5, and 7**: Defender response outcomes, encoded web attacks, Windows password spraying, and suspicious sensitive-file access. They turn common imported evidence into useful findings while directly addressing overstatement risks illustrated by the source.

### 1. Defender detection → response outcome → payload retry or execution

**Analyst value:** Show whether malware was detected only, blocked, allowed, removed, or unsuccessfully remediated, and flag a subsequent execution attempt or observed execution of the same payload.

- Interpret Defender Operational events 1116–1119 using the action, status, errors, threat identity, affected resource, and event time. An action event is not automatically a successful cleanup; an Allow action must not appear as remediation.
- Correlate a malicious file found inside a mounted image with a later copy and process start on the same host. Prefer matching hashes and explicit file/process relationships; path-only links have lower confidence. Separate the container hash from the embedded executable hash.
- Present a short evidence chain with the prevention result, retry, observed process start, and any attributable follow-on activity. A process start supports execution, not completion of the payload's objective.
- No second antivirus alert is **not** evidence that execution succeeded. A scanner running as SYSTEM does not identify the person who launched the payload. A drive letter alone does not identify removable media.

**Current gap:** Command-based Defender tampering is covered; a dedicated imported detection/remediation/retry sequence was not found in the reviewed paths.

**Source:** Incident Detection and Identification → Passive Incident Alerting; Identifying Attack Chains.

### 2. Phishing campaign scope with a separate outcome for every recipient and host

[...]

PR #960, opened 2026-09-11
fix(spill): a masked value elsewhere on the line no longer hides a real secret

hasamba:
## What changed

`secretSpillSignal` judged masking per LINE: one `MASKED` test over the whole text ran before any rule, so a `note=REDACTED`, a `****` or a `......` anywhere on the line silenced every credential rule for that line. A real JWT beside a redaction note, or a real AWS key after a vendor's masked example, graded Info — and Info never reaches the forensic timeline, so synthesis never saw the spill.

Masking is now judged per occurrence: every occurrence of each rule is inspected until one is real or the text is exhausted. A masked-shaped occurrence is skipped (still no Medium manufactured from a redacted value); it no longer hides a real token later on the same line. No occurrence cap — a cap of N is exactly N decoys before the real key. The global regex is built per call, so `lastIndex` can never leak between lines.

Six callers (bash history, syslog, combined web log, ECAR, SIEM, tradecraft) share the table; none changes.

## Second commit (code review)

Codex's code review found the per-occurrence scan still missed two shapes: the generic value class runs over `&`, `=` and `,`, so `TOKEN=xxxxxxxxxxxx&TOKEN=<real>` was ONE match whose masked head hid the real tail; and the database rule judged masking across username and password together, so a redacted username hid a real password. Masking is now judged on the captured value where a rule has one (the `db_uri` password is captured), and a rejected occurrence resumes the scan one character past its start, never past its end. Both pinned by tests.

## Why now

Found while designing #930 item 3 (decoded request targets reuse this table). Landing it alone so the reviewer sees the one function and its tests.

## Tests

- `secretSpillRules.test.ts`: a real JWT beside `note=REDACTED`; the real generic secret after a masked one; a real vendor key after forty masked-shaped decoys (and the decoys alone still null); two calls in a row do not share scan position. Three of the four fail at 32f48910.
- `combinedLogImport.test.ts`: the Referer spill line with `?session=REDACTED` on the path still grades Medium + T1552.001 through the importer.
- The existing "already masked or redacted" cases keep passing unchanged.

part of #930


hasamba:
Checked the follow-up before taking it. The boundary `(?:^|[^A-Za-z0-9])x{6,}` does not do what the note says:

- `AKIA` + sixteen `X` → **not** masked any more (the `A` before the run is alphanumeric), so a vendor's own masked example would be flagged as a real key — the case the new test pins.
- `ghp_xxxxxx<suffix>` → still masked (`_` is a boundary), so the interior-run false negative is not fixed either.

Probed with node on the three shapes. The interior-run false negative is real and pre-existing; it needs a rule-aware check (masked-ness of the key material after the vendor prefix), not a boundary on the whole-line pattern. Leaving it out of this PR.

PR #1007, opened 2026-09-13
fix(super-timeline): a starred or tagged raw row survives the cap (#958)

### LINKED ISSUE #958: Super-timeline cap evicts starred and tagged rows; the star outlives the evidence
## What happens

The super-timeline cap (`DFIR_SUPERTIMELINE_MAX`, default 100,000) evicts rows without looking at whether the analyst starred or tagged them. Today eviction goes by event time (`caseSqliteWorker.ts` `writeSuperEvents`: `ORDER BY coalesce(timestamp_ms, -9007199254740991), row_id`), so a starred row from the oldest part of the history is the first to go once enough newer rows arrive. Afterwards `promoteSuperTimeline` (`routes/timeline.ts` ~254-259) and `starredReport` (`ai/viewReports.ts` ~135-148) look the id up, find nothing, and silently skip it — the star survives in the tags store, the evidence it pointed at is gone, and nothing tells the analyst.

## Why the obvious fix is wrong

The worker has a `super_labels` table, but it is the LEGACY sidecar. Stars and analyst tags are written to `TagsStore` (`routes/timeline.ts` ~154-166 builds the label map from `tagsStore`; `routes/aiSynthesis.ts` ~404-408 reads stars from it). An eviction exemption keyed on `super_labels` protects nothing.

## Acceptance criteria (from the adversarial design review of #932 item 12, round 4)

1. One authoritative, transactionally visible protection relation inside the SQLite worker, fed from `TagsStore` mutations by analyst authors only (`tagger:` authors excluded — tagger tags are automatic and can cover most rows).
2. `setSuperLabels`/the sync validates that the target row exists; a label on an evicted or unknown id is refused or reported, not stored as an orphan.
3. Migration (`migrateSuper`) backfills protection BEFORE the cap is enforced; a labelled id that content-dedup aliases to a retained row transfers its protection to that row.
4. Eviction excess is computed from the count of UNPROTECTED rows (`max(0, unprotected - cap)`), and `entity_counts` stores the exact post-delete total. Document the overflow policy: the cap bounds unprotected rows; every protected row is kept.
5. Tests go through the public `/tags` route, not `SuperTimelineStore.setLabels`: star row 0, bulk-import past the cap out of date order → row 0 survives, `promoteSuperTimeline` still resolves it; migration with a labelled id skipped by dedup; a label-versus-append race.
6. `appendSuper` returns the number of rows RETAINED after eviction (0 when the batch was entirely deduped), never the pre-eviction insert count.

## Related

Surfaced while designing #932 item 12 (undated LEAPP rows), which changes eviction to insertion order for a different reason (an undated-rows class must not be able to pin the cap) and explicitly does not address protection. Relates to #956 and #957.

hasamba:
Closes #958.

## What the analyst sees

A row they starred or tagged in the super-timeline stays there, whatever the row count. The cap (`DFIR_SUPERTIMELINE_MAX`, default 100,000) now bounds only unprotected rows; every protected row is kept. Existing cases pick this up on first open: protection is backfilled from the case's tags once. A star placed on a row that is already gone still saves, but protects nothing — it never leaves an orphan.

## Acceptance criteria

| # | Criterion | Where |
|---|---|---|
| 1 | One authoritative relation in the SQLite worker, fed by TagsStore from analyst authors only | `super_protected` table; `TagsStore.add/remove/removeByAuthorPrefix` → `protect/unprotect`; `tagger:` authors skipped |
| 2 | A label on an evicted or unknown id is refused or reported, never stored as an orphan | `protectSuperRows` writes only for an existing row and returns false; `setSuperLabels` returns false → route 404 |
| 3 | Migration backfills before the cap; a dedup-aliased id transfers protection | `migrateSuper` passes the tags file into `writeSuperEvents`, which protects before evicting; alias through `content_key`; `super_protected_synced` flag backfills pre-#958 cases |
| 4 | Excess = `max(0, unprotected − cap)`; `entity_counts` exact | `writeSuperEvents`; the retention comment in `superTimelineStore.ts` documents the overflow policy |
| 5 | Tests through `/tags` | `tests/server/superTimelineProtection.test.ts` — star row 0, bulk import past the cap out of date order, promote resolves it; tagger tag does not protect; unstar releases; star-vs-append race invariant; legacy label route 404. Migration cases in `tests/analysis/superTimelineProtection.test.ts` |
| 6 | `appendSuper` returns rows retained | already true since #962; unchanged |

## Structure

- The super-timeline half of the SQLite worker moves to `caseSqliteWorkerSuper.ts` (spliced into the worker source as text) to stay under the 800-line gate.
- `TAGGER_AUTHOR_PREFIX` moves down to `superTimeline.ts`; `taggerRun.ts` re-exports it. `check:boundaries` is unchanged at 38.

## Verification

`npm run build`, `typecheck`, `lint`, `format:check`, `check:size`, `check:imports`, `check:boundaries` green; full `npm test`: 820 files, 13886 tests passed.


hasamba:
Codex review (two P2s) addressed in the second commit:

- **Cross-file consistency** — the one-shot sync flag is gone. The worker stores the tags file's size + mtime and, whenever they differ from the last sync (restore, crash between the two writes, a case indexed before this PR), re-derives the exact analyst protection set from the file, releases the rest and enforces the cap. Test: "a tags file changed outside TagsStore (a restore) is reconciled on the next store call".
- **Cap on release** — `enforceSuperCap` is one shared step that runs after every append, unprotect and sync, so the unprotected population never rests above the cap. Test: "unprotect enforces the cap at once".

PR #1274, opened 2026-09-18
fix(analysis): normalize network.source.address to match hostBinding's canonical form

### LINKED ISSUE #1187: [Suggestion] Normalize network.source.address (e.g. via hostBinding.canonicalIp) when importers write it
**File:** companion/src/analysis/combinedLogImport.ts:491 (and other writers of canonical.network.source.address)

Both writers the proxy→workstation join trusts (webChainRows.ts for Zeek, combinedLogImport.ts for Squid/combined-log) stamp network.source.address verbatim from the log line. hostBinding.ts's lookup canonicalizes via canonicalIp() (lowercase, IPv6 compression, ::ffff:v4 folding), so a common-but-legal spelling difference splits what should be the same binding: e.g. a Zeek row logging '::FFFF:10.0.0.5' or an uppercase IPv6 will not collide with a 4624's dotted-quad form, producing a false 'no-match' from the new /cases/:id/proxy-host-identity-matches route rather than ambiguity.

Not a merge blocker: the join never misidentifies — it under-matches (reported honestly as no-match).

**Severity:** low — miss-only, no wrong attribution.
**Recommended fix:** call canonicalIp() at the import seam for network.source.address (combinedLogImport.ts:491, webChainRows.ts:431), or note in proxyWorkstationChain.ts that differently-spelled-equivalent IPs can under-match.

Found while reviewing PR #1185.

hasamba:
closes #1187

Both writers the proxy→workstation join trusts (`webChainRows.ts` for Zeek, `combinedLogImport.ts`
for Squid/combined-log) stamped `network.source.address` verbatim from the log line. hostBinding.ts's
own lookup canonicalizes via `canonicalIp()` (lowercase, IPv6 compression, `::ffff:v4` folding), so a
common-but-legal spelling difference (e.g. an uppercase or non-compressed IPv6 form) split what
should be the same binding — a false "no-match" rather than ambiguity. Applied `canonicalIp()` at
both import seams, only to the trust-sensitive, join-relevant `network.source.address` field —
`srcIp`/`actor.address` (general-purpose display fields) are deliberately left at the line's own raw
spelling.

One Ollama code review round (verdict: needs-attention, all three findings verified false/out-of-
scope by direct evidence, none applied):
- **"No backfill for pre-existing envelopes"** — rejected: no other fix in this codebase, or
  anywhere else in this session's own backlog, backfills historical case data; forensic re-import is
  the established remedy, not a background migration.
- **"New divergence between the canonical field and sibling srcIp/actor.address"** — rejected:
  grepped every real consumer of `srcIp` across the codebase; none perform an equality comparison
  against `canonical.network.source.address` — several deliberately prefer-fallback-chain (canonical
  first, `srcIp` only if canonical is absent), unaffected by value divergence.
- **"canonicalIp() could throw or mishandle adversarial input"** — rejected: `canonicalIp()`'s only
  fallible branch is wrapped in try/catch (provably throw-safe for any string); zone-scoped IPv6 is
  already stripped before the check; directly verified via `node -e` that Node's real `isIP()`
  already rejects the two specific examples raised (a bracketed literal, a leading-zero IPv4 octet)
  before they ever reach `canonicalIp()` through either writer's own upstream validity gate.

Gate pass: build/typecheck/lint/format:check, all four `check:*` gates (one expected drift —
ARCHITECTURE.md's own comply/total pair, re-baselined via `check-boundaries.mjs --update` after the
new `hostBinding.js` imports), full `npm test` (966 files / 16798 tests) all green. Secret scan of
the diff clean.


hasamba:
Closing without merging: #1187 was already correctly resolved by a concurrent session (closed as not-reproducible — verified that `resolveIpAtTime` already calls `canonicalIp()` on its own query-side lookup key, so both sides of the IP comparison are normalized at read time regardless of the stored value's own spelling). I independently confirmed this against the real `resolveIpAtTime` source after the fact — my own diagnosis mis-applied a pattern from a different, unrelated normalization gap (#1160's IP-exclusion logic) without first tracing whether the read-side lookup already normalized. This PR's own write-time normalization is harmless in isolation but solves a problem that doesn't actually exist in the real join, and duplicates/conflicts with the already-merged, correct resolution. No action needed on #1187 (already closed).

PR #1324, opened 2026-09-18
feat(analysis): provenance-tag IOCs read from client-forgeable fields — client-reported never feeds an acting block-list

### LINKED ISSUE #1266: feat(analysis): provenance-tag IOCs derived from client-forgeable fields (email originatingIp and similar)
**Surfaced by**: #1184's own Ollama code review (a real, pre-existing gap that predates the
branch — not something #1184's own diff introduced or worsened).

**What's missing**: #1184 stopped stamping emailImport.ts's `originatingIp` (X-Originating-IP /
earliest external Received hop — can be attacker-influenced) onto the trust-sensitive
`canonical.network.source.address` field, since a host-identity join treats that field as
edge-observed. `originatingIp` is still added as a plain `"ip"`-type IOC
(`addIoc(sink, "ip", p.originatingIp)`, pre-existing behavior, unchanged by #1184) with no
provenance marker distinguishing it from an IOC extracted from a genuinely edge-observed source
(e.g. a Zeek connection record's own real peer address). An analyst pivoting on IOCs, an
allowlist/blocklist enforcement pass, or an automated export could treat a forged
X-Originating-IP the same as a verified network fact.

**What would close this**: either (a) a confidence/provenance field on the IOC record itself
(e.g. `confidence: "verified" | "client-reported"`), populated at every extraction site that reads
from a client-forgeable field, with downstream consumers (enforcement/export/blocklist paths)
excluding or flagging low-confidence IOCs, or (b) at minimum, a documented, deliberate scope
decision that IOC extraction is analyst-facing-only (pivots, not automated action) and no
enforcement path currently consumes IOCs unmarked — confirm which is actually true before
building anything.

**Scope note**: this is a change to the shared IOC record shape and every extraction site, not a
one-file fix — real, but a different and larger surface than #1184's own writer-only audit of
`canonical.network.source.address`. Not attempted in #1184 itself.

hasamba:
closes #1266

**The concrete harm, verified against real source**: `mispPush.ts` sets `to_ids: true` (MISP's
automated-IDS/blocking flag) on every ip/domain/hash/url IOC, and `reports/iocBlocklist.ts` is a
literal acting block-list export. So a forged `X-Originating-IP` was shipped as a blockable
network indicator, indistinguishable from a sensor-observed peer. The issue's fallback ("confirm
no enforcement path consumes IOCs unmarked") is false; option (a) landed.

**Mechanism**: one asymmetric literal `provenance?: "client-reported"` on `SiemIoc`/`IOC`, set at
exactly one site today — `emailImport.ts`'s `originatingIp`. Absent is NOT "verified"; it is
"extracted the ordinary way", every IOC's state before this field. The lure's own URL/host/
attachment IOCs stay unmarked on purpose (content vs attribution — blocking a lure host is the
point; blocking a claimed relay on the claim alone could block a legitimate MTA).

**Merge lattice**: plain wins at every sink (`addIoc`, `mergeRowIocs`, `stateMerge`, `iocMerge`).
An existing marker is cleared only by a later ordinary sighting that carries `extractedFrom` — the
one signal every model delta already has stripped at `stripAiExtractedFrom` — so a model can
neither mark an IOC nor un-mark one by restating it.

**Consumers**: block-list (txt/csv/stix) excludes marked values outright and says so in its
header; MISP push `to_ids: false` + explanatory comment; IRIS description-first + `client-reported`
tag; report IOC table and geo table suffix the value. Enrichment, provenance-chain and the
synthesis prompt deliberately unchanged (see `RECOMMENDATION-1266.md`).

**Real defect found by the end-to-end test the code review asked for**: `platformImports.ts`'s
email delta builder whitelisted `{ id, type, value }` and dropped the marker — it never reached
`mergeDelta` in production while every unit test passed. Fixed; the new
`iocClientReportedEndToEnd.test.ts` runs a real `.eml` through `pipeline.importEmail`, a save/load
round-trip, both block-list formats, and a later Zeek conn sighting that clears the marker — and
was confirmed red with the fix reverted.

**Also**: the IOC-sink section (`addIoc`/`mergeRowIocs`/`resolveExtractedFrom`/`SiemIoc`) moved
verbatim from `siemImport.ts` into `iocSink.ts` for the file-size ledger, re-exported so every
importer is unchanged; one unrelated comment merge in `markdown.ts` for the same ledger.

**Reviews** (Ollama, glm-5.3:cloud): design round "accept with fixes" — HIGH (model restatement
would clear the marker) fixed via the `extractedFrom` gate, plus 2 MEDIUM/8 LOW applied; code round
"accept with fixes", no High — MEDIUM-1/2, LOW-4/5/6/7, NIT-9 and all named test gaps applied;
LOW-3 (a marker can outlive unlinked telemetry, fails toward under-blocking) and NIT-10 (already-
pushed MISP attributes keep `to_ids`) accepted and disclosed.

Gate pass: build/typecheck/lint/format:check, all four `check:*` gates, full `npm test`
(971 files / 17043 tests) after merging master. Secret scan of the diff clean.


hasamba:
Heads-up from a `/gha-triage` run: two commits were stacked onto this branch (`4b0de259` → `7ca39280`) fixing follow-up issues that depend on the `client-reported` field this PR introduces:

- `4337bf83` — #1325: the full STIX bundle now marks a client-reported IOC (`labels: ["client-reported"]` + the same description line `irisMap.ts` uses). Blocklist STIX unchanged.
- `7ca39280` — #1326: dashboard IOC row chip, geo popup/list suffix, and Notion IOC table suffix — the same `(false positive)` pattern. `public/` files edited in place, not reformatted.

Both were built in isolated worktrees off this branch, RED→GREEN, and re-verified on top of the current head (263 targeted tests + typecheck green). Your local checkout will fast-forward on the next pull. When this PR merges, #1325 and #1326 can be closed — they are not in this PR body's close list.

